PR #7, opened 2018-04-26
feat(api): implement a basic /send endpoint

### LINKED ISSUE #1: Implement the basic /send API
`POST /send`, payload:

```
{
  to,      // Array[strings], required. Email addresses of main recipient(s).
  cc,      // Array[strings], optional. Email addresses of cc recipient(s).
  subject, // String, required. The subject of the email.
  body: {
    text,  // String, required. The plaintext body.
    html   // String, optional. The HTML body.
  }
}
```

philbooth:
Fixes #1.

Not wired in to any behaviour, just does some crude validation and then returns a hard-coded message id. There is some test coverage, but also some gaps where I haven't figured out how we should do stuff yet:

* The validation code is punishingly verbose. I evauated a couple of crates to implement validation, `accord` and `validator`, and opted for `validator`. However they both seem quite simple, for instance there's no support for nested objects or values packed inside arrays. This meant we couldn't lean on the nice, readable macro syntax that's available and have to explitly validate fields using Rocket's `FromData` trait instead.

* Rocket has a feature called error catchers, which I couldn't get to work. These are supposed to let you return custom error data but, no matter what I tried, it always seemed to return the default error response for each type (which is an html string). And even if I could get those to work, I couldn't see a way to propagate rich error information, e.g. names of invalid parameters, into the catcher without having to validate all of the parameters a second time inside the catcher.

* I tried to write a test case that fired unicode at the endpoint and it failed.

I'm sure the root cause of these problems is just my own lack of expertise and we'll figure it out eventually. In the meantime, this seemed like enough to get us started. I'll open issues to cover all of the above.

In terms of code structure, I've just lumped most stuff in to the `send` module for now. As we add more functionality, I expect the natural boundaries to reveal themselves. No point agonising over it yet.

@rfk, @mozilla/fxa-devs r?

### DIFF
--- a/src/send/mod.rs
+++ b/src/send/mod.rs
@@ -1,0 +1,100 @@
+use std::{borrow::Cow, collections::HashMap};
+
+use rocket::{
+  data::{self, FromData},
+  http::Status,
+  Data,
+  Outcome,
+  Request,
+};
+use rocket_contrib::{Json, Value};
+use validator::{self, ValidationError};
+
+#[cfg(test)]
+mod test;
+
+#[derive(Debug, Deserialize)]
+struct Body
+{
+  text: String,
+  html: Option<String>,
+}
+
+#[derive(Debug, Deserialize)]
+struct Email
+{
+  to: Vec<String>,
+  cc: Option<Vec<String>>,
+  subject: String,
+  body: Body,
+}
+
+impl FromData for Email
+{
+  type Error = ValidationError;
+
+  fn from_data(request: &Request, data: Data) -> data::Outcome<Self, Self::Error>
+  {
+    let result = Json::<Email>::from_data(request, data);
+    match result
+    {
+      Outcome::Success(json) =>
+      {
+        let email = json.into_inner();
+
+        if email.to.len() == 0
+        {
+          return fail();
+        }
+
+        if let Some(outcome) = validate_addresses(&email.to)
+        {
+          return outcome;
+        }
+
+        if let Some(ref cc) = email.cc
+        {
+          if let Some(outcome) = validate_addresses(&cc)
+          {
+            return outcome;
+          }
+        }
+
+        Outcome::Success(email)
+      },
+      Outcome::Failure(_error) => fail(),
+      Outcome::Forward(forward) => Outcome::Forward(forward),
+    }
+  }
+}
+
+fn fail() -> data::Outcome<Email, ValidationError>
+{
+  Outcome::Failure((
+    Status::BadRequest,
+    ValidationError {
+      code: Cow::from("400"),
+      message: None,
+      params: HashMap::new(),
+    },
+  ))
+}
+
+fn validate_addresses(addresses: &[String]) -> Option<data::Outcome<Email, ValidationError>>
+{
+  for address in addresses
+  {
+    if !validator::validate_email(&address)
+    {
+      return Some(fail());
+    }
+  }
+
+  None
+}
+
+#[post("/send", format = "application/json", data = "<_email>")]
+fn handler(_email: Email) -> Json<Value>
+{
+  Json(json!({ "messageId": 0 }))
+}
--- a/src/send/mod.rs
+++ b/src/send/mod.rs
@@ -8,7 +8,7 @@ use rocket::{
   Request,
 };
 use rocket_contrib::{Json, Value};
-use validator::{self, ValidationError};
+use validator::{self, Validate, ValidationError};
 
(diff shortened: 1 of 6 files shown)


src/send/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+struct Body
+{
+  text: String,
+  html: Option<String>,
+}
+
+#[derive(Debug, Deserialize)]
+struct Email
+{
+  to: Vec<String>,
```
rfk: I'm surprised this is a `Vec` rather than a singe `String`, do we sometimes send multiple emails in the `To:` field from auth-server?

src/send/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        let email = json.into_inner();
+
+        if email.to.len() == 0
+        {
+          return fail();
+        }
+
+        if let Some(outcome) = validate_addresses(&email.to)
+        {
+          return outcome;
```
rfk: When I first read this I was a little confused, become `outcome` sounds a bit like is succeeded.  Would it read better as `Some(error)` or similar?

src/send/test.rs
```diff
@@ -0,0 +1,222 @@
...
+  self,
+  http::{ContentType, Status},
+  local::Client,
+};
+
+fn setup() -> Client
+{
+  let server = rocket::ignite().mount("/", routes![super::handler]);
+
+  Client::new(server).unwrap()
```
rfk: This is pretty neat!

philbooth:
> ...we could ask Ben Bangert or some of his team for some input here as well...

Yep, the more the merrier!

src/send/mod.rs
philbooth: Nope we don't, it's just me failing at YAGNI. And this will make the validation a bit nicer too fwiw, so win win.

src/send/mod.rs
philbooth: Tbh, all the naming here dissatisfies me. `validate_addresses` is misleading too, given the implementation. I think I'm better at naming things in the a.m., will sleep on it! 😄

src/send/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        if let Some(ref cc) = email.cc
+        {
+          if let Some(outcome) = validate_addresses(&cc)
+          {
+            return outcome;
+          }
+        }
+
+        Outcome::Success(email)
```
bbangert: This whole block could use the function chaining a bit.
```rust
email.cc
    .map(|ref cc| { validate_addresses(&cc) })
    .unwrap_or(Outcome::Success(email))
```

bbangert: It's also possible Rust might not like my shorter version, since it might still be borrowing cc from email... the joy of lifetimes.

src/send/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+  body: Body,
+}
+
+impl FromData for Email
+{
+  type Error = ValidationError;
+
+  fn from_data(request: &Request, data: Data) -> data::Outcome<Self, Self::Error>
+  {
+    let result = Json::<Email>::from_data(request, data);
```
pjenvey: I recommend avoiding the Outcome API as much as possible. The code below and fns like validate_address will be more straightforward built around the basic Result/Option types (and maybe easier to get it looking like bbangert's suggestion below).

Defer converting to Outcome until as late as possible. I've gone as far as having the request guard call out to a separate function returning probably a Result, then doing a final conversion

src/send/mod.rs
```diff
@@ -0,0 +1,99 @@
+#[cfg(test)]
+mod test;
+
+use rocket::{
```
pjenvey: Good practice to arrange these according to this Style guide: https://doc.rust-lang.org/1.0.0/style/style/imports.html

pjenvey: and nitpick: Personally I'd stick w/ the default cargo fmt options, I rarely see all of these :)

src/send/test.rs
```diff
@@ -0,0 +1,222 @@
...
+      "body": {
+        "text": "baz"
+      }
+    }"#,
+    )
+    .dispatch();
+
+  assert_eq!(response.status(), Status::Ok);
+
+  let body = json!(response.body().unwrap().into_string().unwrap());
```
pjenvey: probably don't need json!() here? Handy for the following line, though:

` assert_eq(body, json!({"messageId": 0}).to_string());`

PR #14, opened 2018-05-03
refactor(settings): extract types for deserialization

philbooth:
Fixes #13.

Extracts the inline validation code from the `settings` module to a dedicated `deserialize` module, so that the validation rules can be better re-used elsewhere.

Relies on a currently-unmerged PR in order to work, hence the `config` crate is now resolved with a git URL rather than a version string. If that situation persists, we can change it to point to a fork under our own control but hopefully it will be merged soon and we can revert to a version string. There's also an equivalent refactoring in the [`pb/deserialize-without-types` branch](https://github.com/mozilla/fxa-email-service/compare/pb/deserialize-without-types?expand=1), which doesn't require this fix but uses serde's `deserialize_with` macro instead. I'm happy to pursue that approach if it's preferable.

The other gotcha was that I had to flatten the `Settings` struct in order for the tests to be able to create clean instances with fresh values. This is a little bit annoying and is not necessary for the alternative, function-based refactoring. I don't have a root cause yet, but I'll open an issue in the appropriate repo as soon as I have a minimal test case (or come back and fix it if something's wrong here, which is eminently possible).

@mozilla/fxa-devs r?

Issue #9, opened 2018-04-27
Ensure unicode handling is sane


philbooth:
Depends on #21. Write tests for unicode characters and make sure they pass.

philbooth:
Part of this will include setting `charset` in `src/providers/ses.rs`, as per https://github.com/mozilla/fxa-email-service/pull/17#discussion_r186314207.

philbooth:
Was going to pick this up then I realised there's not much we can do until we have better tests for SES/Sendgrid. Will come back to it.

brizental:
@philbooth hey, Phil! I'm want to start working on this, can you just explain it to me a little further so that I understand it completely?

vladikoff:
@philbooth should we grab the unicode test cases from the fxa-auth-server? any others?

philbooth:
Yep, there might not be much to it tbh. Essentially all I was thinking was that we don't have any tests that fire unicode characters through our providers yet. Maybe everything is fine, or maybe it isn't.

So at the very least we want a test in `src/send/test.rs` that fires a selection of unicode characters through some of the params to the mock provider. Ideally we'd also have something that stress-tests SES too via the mock SES stuff you did. Then if the new tests pass there's nothing more to do than that!

Make sure you pick some unicode characters from the higher planes, so that we're throwing multi-byte input at it. (the 🦀 emoji might be appropriate there!)

And make sure all the params are covered of course.

philbooth:
> should we grab the unicode test cases from the fxa-auth-server?

Didn't know there were any! If there's some we can re-use then yep!

brizental:
Just documenting my findings here, I think we are actually not able to handle unicode character the way we do things now. I tried this test and it failed: 

```
fn unicode_email_field() {
    let client = setup();

    let random_unicode_string :String = (0..10).map(|_| random::<char>()).collect();
    // let random_unicode_string = String::from("一丁丂七丄丅丆万丈三上");
    let mut response = client
        .post("/send")
        .header(ContentType::JSON)
        .body(format!("{{
            \"to\": [ \"{}@example.com\" ],
            \"cc\": [ \"bar\" ],
            \"subject\": \"baz\",
            \"body\": {{
                \"text\": \"qux\"
            }},
            \"provider\": \"mock\"
        }}", random_unicode_string))
        .dispatch();

    assert_eq!(response.status(), Status::Ok);

    let body = response.body().unwrap().into_string().unwrap();
[...]

Issue #6, opened 2018-04-25
Support an alternative email provider


philbooth:
Keeps us honest in our objective to decouple from SES. Will need equivalent logic for sending and dealing with bounces/delivery notifications.

philbooth:
Possibly of interest: https://crates.io/crates/sendgrid

philbooth:
To do: think about approaches to bounce-handling ahead of time here, what would e.g. a sendgrid bounce handler that works with whatever we do for #5 look like?

Ref: https://sendgrid.com/docs/API_Reference/Webhooks/event.html

philbooth:
I can also imagine something stateful that polls the bounce and spam APIs:

* https://sendgrid.com/docs/API_Reference/Web_API_v3/bounces.html
* https://sendgrid.com/docs/API_Reference/Web_API_v3/spam_reports.html

Stateful because it would need to set an appropriate `start_time` per iteration, which survives restarts etc.

One benefit to this kind of approach is that polling the API fits in conceptually with the SES queue-handling stuff.

A downside to it is that I haven't seen a way to get delivery events yet.

PR #17, opened 2018-05-04
feat(ses): implement ses-based email sending

### LINKED ISSUE #2: Implement basic email sending via SES
Possibly using Lettre. Will also require adding config for connection params etc.

philbooth:
Fixes #2.

To test it out:

```
curl -d '{"to":"foo@example.com","subject":"bar","body":{"text":"baz"}}' -H 'Content-Type: application/json' http://localhost:8000/send
```

~~Note that we're currently pinned to the `nightly-2018-04-28` toolchain while we wait on a Rocket update, so you'll need to set that up like so:~~

```
rustup toolchain install nightly-2018-04-28
rustup override set nightly-2018-04-28
```

Also note that using a simple `cargo t` to run the tests will fail. You need to run them in a single thread:

```
cargo t -- --test-threads=1
```

I added a shell script called `t` to make this less annoying, so you just need to type `./t` to run them.

There's a bunch of implementation decisions that I'll call out in the inline comments.

@mozilla/fxa-devs r?


src/deserialize.rs
```diff
@@ -0,0 +1,69 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
philbooth: This module moved up to the top level because the serde trait and lifetime stuff makes it a pain to test directly. It's tested indirectly via the `settings` module though, and the underlying `validate` module also has tests.

config/default.json
```diff
@@ -1,5 +1,9 @@
 {
+  "provider": "ses",
```
philbooth: This is the default provider but can be overridden per request. That allows the tests to specify a "mock" provider and will allow e.g. local dev to use the "smtp" provider (which doesn't exist yet).

philbooth: (...and for #6 maybe we'll add a "sendgrid" provider)

src/providers/ses.rs
```diff
@@ -0,0 +1,89 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
philbooth: There are no tests for this SES functionality yet, because I haven't worked out how to mock all Rusoto stuff. Hopefully it's okay to get round to that later, but in the meantime you should be able to find an email in your inbox with the subject "fxa-email-service test" that shows it works.

src/providers/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+{
+  fn send(
+    &self,
+    to: String,
+    cc: Vec<String>,
+    subject: String,
+    body_text: String,
+    body_html: Option<String>,
+  ) -> Result<String, ProviderError>;
+}
```
philbooth: Obviously the idea behind this trait is to localise all provider-specific stuff inside the appropriate module and just expose a generic interface to the rest of the code.

src/providers/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+  fn fmt(&self, formatter: &mut Formatter) -> fmt::Result
+  {
+    write!(formatter, "{}", self.description)
+  }
+}
+
+pub struct Providers
+{
+  settings: Settings,
+  providers: HashMap<String, Box<Provider>>,
```
philbooth: And this map is used to direct requests to the appropriate provider, so all the caller needs to know is the provider name ("ses" or "smtp" or whatever).

src/providers/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+  {
+    let resolved_provider_id = if let Some(ref id) = provider_id {
+      id
+    } else {
+      &self.settings.provider
+    };
+
+    match self.providers.get(resolved_provider_id) {
+      Some(ref provider) => match provider.send(to, cc, subject, body_text, body_html) {
+        Ok(message_id) => Ok(format!("{}:{}", resolved_provider_id, message_id)),
```
philbooth: Since SES already returns a guaranteed unique message id, for now I'm opting to re-use that and prefix it with the provider id ("ses" in this case). We don't have to keep it that way forever, but at least this lets us postpone the decision about message ids for a while.

src/settings/mod.rs
```diff
@@ -11,6 +11,23 @@ use deserialize;
 #[cfg(test)]
 mod test;
 
+#[derive(Debug, Deserialize)]
+pub struct AwsKeys
+{
+  #[serde(deserialize_with = "deserialize::aws_access")]
+  pub access: String,
+  #[serde(deserialize_with = "deserialize::aws_secret")]
+  pub secret: String,
```
philbooth: I ummed and ahhed about whether to add any validation for these two or not, in case I accidentally prohibit any valid keys. So they are validated, but it's pretty loose.

src/settings/test.rs
```diff
@@ -6,31 +6,23 @@ use std::{
...
   error::Error,
-  sync::{LockResult, Mutex, MutexGuard},
 };
 
 use super::*;
 
-lazy_static! {
-  // HACK: mutex to prevent tests clobbering each other's environment variables
-  static ref ENVIRONMENT_MUTEX: Mutex<bool> = Mutex::new(true);
-}
```
philbooth: Because the tests are now run with `--test-threads=1`, I could delete this silly hack. (it wasn't preventing these tests from interfering with others that depend on `std::env`)

src/validate/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+
+use regex::Regex;
+use rusoto_core::Region;
+
+#[cfg(test)]
+mod test;
+
+lazy_static! {
+  static ref AWS_ACCESS_FORMAT: Regex = Regex::new("^[A-Z0-9]+$").unwrap();
```
philbooth: This *seems* to be uppercase alphanumeric, but I could be wrong.

src/validate/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+  static ref AWS_SECRET_FORMAT: Regex = Regex::new("^[A-Za-z0-9+/=]+$").unwrap();
```
philbooth: This *seems* to be base64, but I could be wrong.

src/validate/test.rs
```diff
@@ -0,0 +1,117 @@
...
+  assert!(validate::aws_region("us-east-2"));
+  assert!(validate::aws_region("us-west-1"));
+  assert!(validate::aws_region("eu-west-1"));
+}
+
+#[test]
+fn invalid_aws_region()
+{
+  assert_eq!(validate::aws_region("us-east-1a"), false);
+  assert_eq!(validate::aws_region("us-east-1 "), false);
```
philbooth: Hopefully these were the kind of tests you were thinking of in https://github.com/mozilla/fxa-email-service/pull/14#discussion_r185982806 @rfk. Splitting out the separate `validate` module made writing them super-easy.

philbooth:
> Note that we're currently pinned to the `nightly-2018-04-28 toolchain` while we wait on a Rocket update...

Rocket got an update and we're back on vanilla nightly now, so you can disregard this note.

src/providers/ses.rs
rfk: > you should be able to find an email in your inbox with the subject "fxa-email-service test" that shows it works.

Can confirm :-)

src/providers/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+      },
+      None => Err(ProviderError::new(format!(
+        "invalid provider '{}'",
+        resolved_provider_id
+      ))),
+    }
+  }
+}
+
+unsafe impl Sync for Providers {}
```
rfk: It's probably my rust n00bness showing, but what's this about?

src/settings/mod.rs
rfk: On the plus side, prohibiting valid keys is something that will fail safely and be detected quickly, so this seems fine to me.

src/send/mod.rs
```diff
@@ -87,8 +96,34 @@ fn fail() -> data::Outcome<Email, ValidationError>
...
+    Some(provider.clone())
+  } else {
+    None
+  };
+
+  match PROVIDERS.send(
+    email.to.clone(),
+    cc,
+    email.subject.clone(),
+    email.body.text.clone(),
```
rfk: Naively, I would expect all of these strings to say alive for the entire duration of the request, and hence to be able to pass borrows rather than clones.  For my own learning purposes, could you please explain why the clones are needed here?

src/providers/ses.rs
```diff
@@ -0,0 +1,89 @@
...
+    let mut destination = Destination::default();
+    destination.to_addresses = Some(vec![to]);
+    if cc.len() > 0 {
+      destination.cc_addresses = Some(cc);
+    }
+
+    let mut message = Message::default();
+    message.subject.data = subject;
+    message.body.text = Some(Content {
+      charset: None,
```
rfk: Should we use an explicit charset here, or does it default to doing something sensible?

src/providers/mod.rs
philbooth: Implementing `Sync` tells everything else that `Providers` is threadsafe. Specifically, we want to store a static ref in `src/send/mod.rs`, hence the need to implement `Sync`.

And obviously I can be confident that it *is* threadsafe because none of the methods are `mut`.

src/providers/ses.rs
philbooth: The relevant docs are these:

https://rusoto.github.io/rusoto/rusoto_ses/struct.Content.html

I have an issue open for unicode support in #9 and there's no immediate need to send unicode-enabled email right now, so I figured it was okay to leave the default behaviour in this PR and come back to implement uncode properly later on, in an everything-definitely-works-end-to-end fashion.

philbooth: (but I'll add a comment to the issue to remind me that this bit needs doing too)

src/send/mod.rs
philbooth: There were a few steps that led me to this point so I'll recount them in turn as maybe I took a wrong step along the way:

1. [Rusoto's `Content` type](https://rusoto.github.io/rusoto/rusoto_ses/struct.Content.html) takes allocated `String` instances, so at some point I was going to have to choose between cloning the strings or passing ownership to Rusoto.

2. Passing ownership to Rusoto would preclude us from using the strings after it's been invoked. Maybe we want them afterwards for logging or some other post-send activity like setting up bounce handlers.

3. At that point the choice is between either cloning inside the SES provider and passing it `&str` or cloning here and passing `String`. To some extent the right decision here depends on what the implementation will look like for other providers. If others also need cloned strings, it makes sense to do that once, outside. If not it makes sense for that to be SES-specific behaviour and pass string slices. I thought about it for a bit, wasn't sure which would be the case and took a punt on cloning outside.

I'm happy to change this to pass ownership of the strings to Rusoto, or to clone them inside the SES provider, whichever you're more comfortable with. I wasn't really sure which would turn out to be right in the long run and optimistically hoped it would all come out in the wash. :smile:

src/providers/mod.rs
philbooth: I've only read a little bit about enums and haven't used them in anger yet, so I'm not quite sure what that implementation would look like in detail.

It seems like we'd still need something in place of the `HashMap` here to do the mapping from the string to the enum value anyway. And I don't know what those enum values would look like. Would we still have the `Provider` trait and different implementations thereof, or would they directly be variants of the enum? Can functions be enum values? (not looking for answers here, just listing some of the things I don't understand yet)

I can play around with it tomorrow though and see where I end up.

src/send/mod.rs
rfk: > I'm happy to change this to pass ownership of the strings to Rusoto, or to clone them inside
> the SES provider, whichever you're more comfortable with.

I'm very happy to leave them as `clone`s here, just trying to learn as I go :-)

IIUC this is a pretty common evolution for rust code, start off with a bunch of clones until you stabilize the interefaces and figure out which can be borrows etc, so 👍 let's not block on getting that all exactly right.

src/providers/mod.rs
rfk: > I can play around with it tomorrow though and see where I end up.

On the other hand, it's probably very easy for us to get lost in the type-system weeds here, I'm very happy for it to stay as a `HashMap` while we focus on fleshing out all the required behaviors for this repo.

src/send/mod.rs
philbooth: Fwiw, after sleeping on this I decided to switch to slices in the `Provider` interface and then do the cloning inside the `Ses` provider as a specific step.

Issue #8, opened 2018-04-27
Implement custom errors


philbooth:
Rocket has a feature called error catchers that I failed to get working in #1. Let's use it and stop returning HTML strings for error payloads.

philbooth:
As part of this work we can also define a proper top-level error type. I've been accumulating ad-hoc error implementations in other modules so far, we should get rid of those (`ProviderError`, `DbError`).

Issue #5, opened 2018-04-25
Handle SES delivery and bounce notifications


philbooth:
Handle them by sending messages to SQS. Message format:

```
{
  messageId, // Number, required.
  states: [
    {
      address,  // String, required. The recipient addre
      state     // String, required. One of: delivered|bounced-hard|bounced-soft|complained
      expiresAt // Number, required.
    }
  ]
}
```

Depends on #4 for `messageId`.

philbooth:
Bounce notifications should also update the `emailBounces` table of course.

philbooth:
Also keep in mind that whatever we do here will need to work well with other providers, sendgrid or whatever. See #6.

PR #22, opened 2018-05-12
feat(db): check the emailBounces table before sending email

### LINKED ISSUE #3: Check the emailBounces table before sending
Bounce limits should be set in config. Appropriate error details should be returned if any recipient address violates the limits:

```
{
  code: 429,
  errno: 103,
  message: 'Recipients have bounced',
  states: [ { address, state, expiresAt } ]
}
```

philbooth:
Fixes #3. (conference got in the way before I could finish it off, sorry for the delay)

Loads bounce limits from config and then checks email addresses against those limits before sending the email. If any address violates a limit, the request will fail with status `429` and email will not be sent.

As it introduces a dependency on the fxa-auth-db-mysql repo, some extra scripting has been thrown in to make working with node.js easier.

Running `./tdb` will check whether an `fxa-auth-db-mysql` directory exists and, if not, clone the repo. Then it will check for running db server processes and, if none are found, run `npm i`, run the db patcher and start the db server in a background process. Then it finishes by invoking the existing `./t` script.

This is not ideal because it leaves the node process running in the background after tests have finished. But I figure it's good enough for now because it only leaves one of them running, they don't multiply with subsequent test runs. And you don't have to use the new script of course, if you already have the db server running in fxa-local-dev, you can just run `./t` and the tests should pass. The new script is mostly there for Travis' benefit.
    
Longer term, if this service is going to stick around, we can do something more permanent to manage it properly.

@mozilla/fxa-devs r?


src/auth_db/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl Display for DbError
+{
+  fn fmt(&self, formatter: &mut Formatter) -> fmt::Result
+  {
+    write!(formatter, "{}", self.description)
+  }
+}
+
+impl From<UrlError> for DbError
```
philbooth: This `From` trait was the crucial missing link in my understanding of how we can implement clean error handling in this repo. We can lean on it for #8.

src/auth_db/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+
+  pub fn get_email_bounces(&self, address: &str) -> Result<Url, UrlError>
+  {
+    self.get_email_bounces.join(&hex::encode(address))
+  }
+}
+
+pub trait Db
+{
+  fn get_email_bounces(&self, address: &str) -> Result<Vec<BounceRecord>, DbError>;
```
philbooth: Putting the db methods in a trait means we can create mock db structs for testing purposes.

src/auth_db/test.rs
```diff
@@ -0,0 +1,26 @@
...
+#[test]
+fn get_email_bounces_invalid_address()
+{
+  let settings = Settings::new().expect("config error");
+  let db = DbClient::new(&settings);
+  match db.get_email_bounces("") {
+    Ok(_) => assert!(false, "DbClient::get_email_bounces should have failed"),
+    Err(error) => assert_eq!(error.description(), "auth db response: 400 Bad Request"),
+  }
+}
```
philbooth: There aren't many test cases here because we only have a method for reading. When we implement #5, we can improve the coverage by asserting that the result from `get_email_bounces` matches what we write with `create_email_bounce`.

src/bounces/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+unsafe impl<'a> Sync for Bounces<'a> {}
+
+fn is_bounce_violation(count: u8, created_at: u64, now: u64, limits: &Vec<BounceLimit>) -> bool
+{
+  for limit in limits.iter() {
+    if count > limit.limit && created_at >= now - limit.period {
+      return true;
```
philbooth: A side-effect of using a data format that maps to strong types is that I find this logic much easier to read than the [equivalent code in the auth server](https://github.com/mozilla/fxa-auth-server/blob/6b58bf9b7c856bfcd9e90f8dcc33e3b0ecaac23f/lib/bounces.js#L58-L69).

src/send/mod.rs
```diff
@@ -97,10 +103,17 @@ fn fail() -> data::Outcome<Email, ValidationError>
 }
 
 #[post("/send", format = "application/json", data = "<email>")]
-fn handler(email: Email) -> Result<Json<Value>, Error>
+fn handler(email: Email) -> Result<Json<Value>, Failure>
```
philbooth: Returning `Error` meant that all failures get treated as `500`. This lets us return a `429`.

src/bounces/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+impl From<BounceError> for Failure
+{
+  fn from(_error: BounceError) -> Failure
+  {
+    // Eventually we should be able to do something richer than this,
+    // as per https://github.com/SergioBenitez/Rocket/issues/586.
+    Failure(Status::TooManyRequests)
```
philbooth: This is probably the wrong layer of abstraction for setting the response status, but we're going to properly sort out the error handling in #8 and I figured this will do in the meantime.

src/validate/mod.rs
```diff
@@ -16,6 +16,9 @@ lazy_static! {
   static ref SENDER_FORMAT: Regex = Regex::new(
     "^[A-Za-z0-9-]+(?: [A-Za-z0-9-]+)* <[a-z0-9-]+@[a-z0-9-]+(?:\\.[a-z0-9-]+)+>$"
   ).unwrap();
+  static ref BASE_URI_FORMAT: Regex = Regex::new(
+    "^https?://[A-Za-z0-9-]+(?:\\.[A-Za-z0-9-]+)*(?::[0-9]+)?/(?:[A-Za-z0-9-]+/)*$"
```
philbooth: This regex is overly-restrictive as a general base URI format, but it's correct for our usage.

rfk:
> As it introduces a dependency on the fxa-auth-db-mysql repo

TBH it never occurred to me that this might talk to fxa-auth-db, I had kinda of just assumed it would talk directly to MySQL.  The split makes sense though, by talking to auth-db we don't have to copy the db credentials into a second process.  I guess eventually we'll want this to have its own db rather than sitting on top of auth-db?

src/bounces/test.rs
```diff
@@ -0,0 +1,189 @@
...
+fn now_as_milliseconds() -> u64
+{
+  let now = SystemTime::now()
+    .duration_since(SystemTime::UNIX_EPOCH)
+    .expect("system time error");
+  now.as_secs() * 1000
+}
+
+#[test]
+fn check_no_bounces()
```
rfk: The auth-server has some interesting test-cases for situations that shouldn't bounce, such as:

  * https://github.com/mozilla/fxa-auth-server/blob/9731a080587f24deaec1155b82d8679abd8a2bc4/test/local/bounces.js#L97
  * https://github.com/mozilla/fxa-auth-server/blob/9731a080587f24deaec1155b82d8679abd8a2bc4/test/local/bounces.js#L122

It's not clear to me whether these are equivalently covered by the tests above, but they're probably worth adding if not.

src/send/mod.rs
rfk: If `429` the appropriate code to return here?  IIUC, it's not that the auth-server itself has sent too many requests, but that there's a problem with the specific request being sent.  I wonder if a `400` and an `errno` would be a better fit here.

src/settings/mod.rs
```diff
@@ -41,6 +65,8 @@ pub struct Smtp
 #[derive(Debug, Deserialize)]
 pub struct Settings
 {
+  pub authdb: AuthDb,
+  pub bouncelimits: BounceLimits,
```
rfk: Is the alllowercase capitalization on these a by-product of the code that automagically populates them from environment variables, or just a stylistic choice?

src/deserialize.rs
```diff
@@ -27,13 +43,55 @@ where
...
+    if let Ok(multiplier) = matches.get(1).map_or(Ok(1), |m| m.as_str().parse::<u64>()) {
+      match matches.get(2).map_or("", |m| m.as_str()) {
+        "second" => Ok(multiplier * SECOND),
+        "minute" => Ok(multiplier * MINUTE),
+        "hour" => Ok(multiplier * HOUR),
+        "day" => Ok(multiplier * DAY),
+        "week" => Ok(multiplier * WEEK),
+        "month" => Ok(multiplier * MONTH),
+        "year" => Ok(multiplier * YEAR),
+        _ => fail::<D>(&value),
```
rfk: I think we should add some explicit testcases for this mapping of time strings to integers.

src/bounces/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+      limits: &settings.bouncelimits,
+    }
+  }
+
+  pub fn check(&self, address: &str) -> Result<(), BounceError>
+  {
+    let bounces = self.db.get_email_bounces(address)?;
+    let now = SystemTime::now()
+      .duration_since(SystemTime::UNIX_EPOCH)
+      .expect("system time error");
```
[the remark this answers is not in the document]
philbooth: Yep, it panics if the argument to `duration_since` is earlier/less than the `SystemTime` instance on which it's invoked. So in this case that means if the system clock has been set to a time before Jan 1st 1970.

Given that our reliance on Unix-time crosses service boundaries and the raw values are physically stored in the database, I can think of no appropriate error handling that would allow us to recover here. Aborting is a feature, but one we can file under "never going to happen" for practical purposes.

Ref: https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.duration_since

src/send/mod.rs
philbooth: I did raise this question in the feature doc fwiw, but didn't get any feedback there. I'll run through some of the considerations again quickly:

* I started off with `400` for this in the feature doc then I noticed SergioBenitez/Rocket#46, which talks about `400` only being appropriate for malformed payloads and `422` being correct for other validation errors. That started me thinking.

* Semantically, I think of `400` as an instruction to the caller that it must change the payload for the request to work. But that's not the case here, only some time needs to pass in order for the exact same payload to work fine if tried again. That does seem similar to `429` semantics, even if it's not a rate-limit in the conventional sense.

* Then I looked at the [MDN docs for `429`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) and the language there framed the error in terms of the user:

  > the user has sent too many requests in a given amount of time

  Which made me think it would be okay to consider this response as per-email address rather than a blanket-wide case for the auth server in general. Also the Amplitude API has been in my head recently and they behave in a similar way, returning `429` for specific values of `device_id`.

* On a philosophical level, doesn't "appropriate" here just mean whatever we choose to document for the API (within reason)? This is the point in my thinking where I opted to define it as `"code": 429, "errno": 103` and resolve the comment thread in the feature doc.

* On a practical level, since the implementation work started, I had some issues getting Rocket's error catchers working properly. I'm sure we'll figure those out easily when we go back to it but, in the meantime, we can't differentiate on `errno` anyway for now. So there's a separate argument for leaving this until we resolve #8, regardless, because the tests are stricter if we can differentiate by HTTP status instead.

None of that is strong disagreement btw, I just wanted to show how I reached this point. What is the case for `400` being the correct response?

src/settings/mod.rs
philbooth: The former. If we insert underscores in the keys here, we can't set them as environment variables unless we use some other separator at that level. (and actually, I'm not sure what the valid separators would even be there)

philbooth:
>  I guess eventually we'll want this to have its own db rather than sitting on top of auth-db?

I think that will definitely be a prerequisite for promoting this to a standalone service (if we get to that point). But as long as it's limited to running on an auth-server box, I don't see the value in duplicating not just the config, as you mention, but also the stuff for invoking stored procedures, marshalling result data, for creating and patching the db, testing everything...

src/send/mod.rs
rfk: > I did raise this question in the feature doc fwiw, but didn't get any feedback there

Apologies for missing it there!

> On a philosophical level, doesn't "appropriate" here just mean whatever
> we choose to document for the API (within reason)?

Yes, yes it does.  I don't have any strong objection to the rationale you set out above.  My one concern would be if, in a future where this is generally available, we want to use 429 to tell the actual application server to back off.  I don't have a compellingly better proposal though, so happy to kick that down the road.

src/bounces/test.rs
philbooth: Yeah, I knew my tests here were insufficient but was lazily avoiding doing the legwork that would make better tests possible.

I need to use the trait trick on `Settings`, so that tests can pass in custom config without relying on the real config that happens to be on disk. Overriding the array items in that part of the config via environment variables works, but it assumes deep knowledge of the resolved config that will fail if `dev.json` or `local.json` override stuff.

This will be useful for testing other stuff too though, so it's good that I finally get round to it!

philbooth: > I need to use the trait trick on `Settings`...

Fwiw it was much easier than that, I just needed to derive `Default` and then assign directly to `bouncelimits` on the result of `Settings::default`.

philbooth:
> I was also able to get this building and running locally and tested against my auth-mysql-db.

That's good to hear, thanks! I realised earlier I haven't written a readme yet, so it may not have been the easiest repo to get started with.

config/default.json
```diff
@@ -1,4 +1,21 @@
...
+  },
+  "bouncelimits": {
+    "enabled": true,
+    "complaint": [
+      { "period": "day", "limit": 0 },
+      { "period": "year", "limit": 1 }
+    ],
+    "hard": [
+      { "period": "day", "limit": 0 },
+      { "period": "year", "limit": 1 }
```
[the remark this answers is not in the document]
philbooth: It turns out prod is using the default bounce limits in the auth server after all, so these match as is. We could still change them though, if we think the defaults are too strict in hindsight?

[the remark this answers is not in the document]
rfk: > It turns out prod is using the default bounce limits in the auth server after all

I was sure it we'd reconfigured it, but you're right, I can't find any indication that the config diverges from defaults in production.  Let's keep it as-is for consistency.

rfk:
Just noting my general 👍 to the period refactor and related tests.

PR #27, opened 2018-05-23
fix(db): add missing fields/values to auth_db::BounceRecord

philbooth:
When I did #22, I didn't pay much attention to the `emailBounces` table because there weren't any rows in my table locally. But it quickly became obvious while implementing #5 that I was missing some columns as well as a valid value for the `BounceType` enum.

This changeset fixes both of those issues and includes a couple of other nearby tweaks taken from the same branch, so that I can bring the size of that PR down a bit (it's a beast).

@mozilla/fxa-devs r?

### DIFF
--- a/src/auth_db/mod.rs
+++ b/src/auth_db/mod.rs
@@ -114,6 +114,7 @@ impl From<RequestError> for DbError {
     }
 }
 
+#[derive(Debug)]
 struct DbUrls {
     get_email_bounces: Url,
 }
@@ -135,6 +136,7 @@ pub trait Db {
     fn get_email_bounces(&self, address: &str) -> Result<Vec<BounceRecord>, DbError>;
 }
 
+#[derive(Debug)]
 pub struct DbClient {
     urls: DbUrls,
     request_client: RequestClient,
--- a/src/auth_db/mod.rs
+++ b/src/auth_db/mod.rs
@@ -8,23 +8,73 @@ use std::{
 
 use hex;
 use reqwest::{Client as RequestClient, Error as RequestError, StatusCode, Url, UrlError};
+use serde::de::{Deserialize, Deserializer, Error as DeserializeError, Unexpected};
 
 use settings::Settings;
 
 #[cfg(test)]
 mod test;
 
-#[derive(Clone, Copy, Debug, Deserialize, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub enum BounceType {
-    Hard = 1,
-    Soft = 2,
-    Complaint = 3,
+    Hard,
+    Soft,
+    Complaint,
+}
+
+impl<'d> Deserialize<'d> for BounceType {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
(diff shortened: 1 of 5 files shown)


src/auth_db/mod.rs
```diff
@@ -8,23 +8,73 @@ use std::{
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'d>,
+    {
+        let value: u8 = Deserialize::deserialize(deserializer)?;
+        match value {
+            0 => {
+                // Set whenever the auth db receives a value it doesn't recognise
+                println!("Mapped default auth db bounce type BounceType::Soft");
+                Ok(BounceType::Soft)
```
philbooth: This match arm is the reason why I couldn't derive `Deserialize` any more. I wanted to ensure we had sane behaviour if we read any zeroes from the db and the safest thing to do seemed like treat it as a soft bounce but log when we're doing so.

Note that this differs from the behaviour in the auth server (see mozilla/fxa-auth-server#2453).

src/auth_db/mod.rs
```diff
@@ -8,23 +8,73 @@ use std::{
...
+    // These are mapped from the equivalent SES BounceSubType values
+    Undetermined = 1,
+    General = 2,
+    NoEmail = 3,
+    Suppressed = 4,
+    MailboxFull = 5,
+    MessageTooLarge = 6,
+    ContentRejected = 7,
+    AttachmentRejected = 8,
+    // Not sure where these come from
```
vbudhram: Found some information one these [here](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#bounce-types).

philbooth: Sorry, probably wasn't clear, this comment is about the values that come after it. They're not from SES (at least they're not mentioned in that doc from what I can see), but they are in the db server and I don't understand why.

philbooth: Oh, my bad, they are there as the complaint types. I was looking for them in a table like the bounce type stuff, should have searched. 😊

I'll get rid of the comment.

src/bounces/mod.rs
```diff
@@ -34,7 +34,7 @@ impl BounceError {
 
         BounceError {
             address: address.to_string(),
-            bounce: Some(bounce),
+            bounce: Some(bounce.clone()),
```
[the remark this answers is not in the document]
rfk: Makes sense.

I'm not asking for a change here, just wondering out load: is it better to pass in a borrowed ref have it cloned locally, or to pass in a clone and take ownership of it?  I don't know if there's a standard rust pattern for approching this.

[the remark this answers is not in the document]
philbooth: > is it better to pass in a borrowed ref have it cloned locally, or to pass in a clone and take ownership of it?

Hmm fwiw, as a rule of thumb, cloning as late as possible makes sense to me I think. Mostly on the basis that it puts the call to `clone` closer to thing that mandates its use, but also partly because it opens the door for potential shortcut logic that doesn't need to pay for the clone (although I'm well into the territory of premature optimisation at that point).

PR #28, opened 2018-05-24
feat(queues): handle SES bounce, complaint and delivery notifications

### LINKED ISSUE #5: Handle SES delivery and bounce notifications

philbooth:
Fixes #5.

Opening as a WIP PR for early feedback, while I continue working on transitioning to futures for the queue-handling code in my local branch (my brain is overloaded for today). Or maybe we want to keep the futures stuff in a separate PR as this one is pretty big already. Either way, bear in mind that all of the queue stuff is synchronous here, but won't be when I'm finished. Polling the queues is not fast and there's three queues to check of course, so the main loop chugs round quite slowly when I run the code in the form you see here.

Anyway, the approach I've taken is to split the code into two separate binaries. The entrypoint for the service, i.e. what used to be `src/main.rs`, is in `src/service_main.rs` and the entrypoint for the queue-handling stuff is in `src/queues_main.rs`. I've also added a couple more shell scripts to save keystrokes when running each binary, `./r` for the service and `./rq` for the queues.

I've made a reasonable effort to keep the SQS stuff behind a trait, but it's by no means a perfect abstraction because the SES `Notification` format leaks out to the main queues module. I figured there's already enough crazy data marshalling going on for one PR here as it is, plus it's easy enough to come back and make that perfect down the line if/when we want to.

A nice bit of strong-typing worth pointing out is the division of `Queue` capabilities into `Incoming` and `Outgoing` traits. This means we can make it compile-time impossible to send messages to queues that we only want to read from and likewise impossible to read from queues that we only want to send to. Maybe not that big a deal in practice, but it seemed like a nice separation of concerns to me.

I'll call out some more of the decisions in the inline comments.

@mozilla/fxa-devs r?


config/default.json
```diff
@@ -2,6 +2,9 @@
   "authdb": {
     "baseuri": "http://127.0.0.1:8000/"
   },
+  "aws": {
+    "region": "us-east-1"
```
philbooth: I moved this to a general `aws` setting on the basis that there's no requirement to use queues in a different AWS region to the one we're sending email from. The auth server's ability to specify different regions for everything seems like something we'd never use here and just makes the config format noisier. But if I'm wrong about that, let me know and I can add separate `ses.region` and `sqs.region` settings.

src/auth_db/mod.rs
```diff
@@ -45,29 +48,99 @@ impl<'d> Deserialize<'d> for BounceType {
...
+        };
+        serializer.serialize_u8(value)
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub enum BounceSubtype {
     // Set by the auth db if an input string is not recognised
-    Unmapped = 0,
+    Unmapped,
```
philbooth: If I'd had the sense to write deserialization tests for `BounceSubtype` in #27, I would have realised then that my derived imlpementation wasn't doing what I thought it was. Hence these changes and explicit implementation of `Deserialize`, which should have been part of that PR.

src/auth_db/mod.rs
```diff
@@ -45,29 +48,99 @@ impl<'d> Deserialize<'d> for BounceType {
...
+            BounceSubtype::Abuse => 9,
+            BounceSubtype::AuthFailure => 10,
+            BounceSubtype::Fraud => 11,
+            BounceSubtype::NotSpam => 12,
+            BounceSubtype::Other => 13,
+            BounceSubtype::Virus => 14,
+        };
+        serializer.serialize_u8(value)
+    }
+}
```
philbooth: Obviously it's a shame how verbose all this data marshalling code is, I'm on the lookout for cunning tricks that will help us reduce it somehow (and let me know if you've got some).

src/auth_db/mod.rs
```diff
@@ -104,36 +177,48 @@ impl Display for DbError {
...
     fn from(error: UrlError) -> DbError {
-        DbError::new(format!("URL error: {}", error.description()))
+        DbError::new(format!("URL error: {:?}", error))
     }
 }
 
 impl From<RequestError> for DbError {
     fn from(error: RequestError) -> DbError {
-        DbError::new(format!("request error: {}", error.description()))
+        DbError::new(format!("request error: {:?}", error))
```
philbooth: For a long time I was debugging errors without realising I could do this to get a half-decent error message out.

src/auth_db/mod.rs
```diff
@@ -162,6 +247,28 @@ impl Db for DbClient {
...
+        let response = self
+            .request_client
+            .post(self.urls.create_bounce())
+            .json(&BounceRecord {
+                address: address.to_string(),
+                bounce_type,
+                bounce_subtype,
+                created_at: 0,
+            })
+            .send()?;
```
philbooth: This is also synchronous of course, but I'm not targeting it as part of the futures stuff I'm doing right now. We can come back to it later but also I'm conscious that long-term the db stuff is earmarked for big changes anyway, so perhaps it's not worth it.

src/queues/mock.rs
```diff
@@ -0,0 +1,98 @@
...
+
+impl<'s> Factory<'s> for Queue<'s> {
+    fn new(id: &'s str, _settings: &Settings) -> Queue<'s> {
+        Queue { id }
+    }
+}
+
+impl<'s> Incoming for Queue<'s> {
+    fn receive(&self) -> Result<Vec<Message>, QueueError> {
+        if self.id == "incoming" {
```
philbooth: It's a bit hackish, but doing this allows the tests to assert that the sanctity of incoming and outgoing queues is preserved.

src/queues/mock.rs
```diff
@@ -0,0 +1,98 @@
...
+                processing_time_millis: 0,
+                recipients: vec![String::from(
+                    "fxa-email-service.queues.mock.complaint@example.com",
+                )],
+                smtp_response: String::from(""),
+                remote_mta_ip: None,
+                reporting_mta: None,
+            });
+
+            Ok(vec![bounce_message, complaint_message, delivery_message])
```
philbooth: One of each type to exercise as much code as possible (the queues module is something of a black box and tricky to interrogate otherwise).

src/queues/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fn process_queue(
+        &self,
+        queue: &Box<Incoming + 's>,
+        handler: &Fn(&Notification) -> Result<(), QueueError>,
+    ) -> Result<(), QueueError> {
+        let messages = queue.receive()?;
+        for message in messages.into_iter() {
+            if message.notification.notification_type != NotificationType::Null {
```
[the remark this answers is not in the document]
philbooth: (the code in `src/queues/sqs.rs` will probably make that more clear)

src/queues/sqs.rs
```diff
@@ -0,0 +1,234 @@
...
+                            self.parse_message(message).unwrap_or_else(|error| {
+                                // At this point any parse errors are message-specific.
+                                // Log them but don't fail the broader call to receive,
+                                // because other messages might be fine.
+                                println!(
+                                    "Queue error receiving from {}: {}",
+                                    self.url,
+                                    error.description()
+                                );
+                                Message::default()
```
philbooth: This is why there's that extra check in `src/queues/mod.rs`.

philbooth:
> ...while I continue working on transitioning to futures for the queue-handling code...

One of the pain points I have with this right now is that rusoto depends on version 0.1 of futures. I wanted to use futures 0.2 for the sweet combinators from `FutureExt`. It took me a while to realise that the reason the compiler was claiming `RusotoFuture` didn't implement the `Future` trait was because I'd imported the wrong version of futures. I guess I'll use version 0.1 instead, but it's a bit of a shame.

philbooth:
I've pushed two variations of the futures stuff. The one you can see here is my preferred option because it simplifies a load of the lifetime stuff, allows for some nice type aliasing and is less verbose, but comes at the (very tiny) cost of duplicating the strings for queue URLs rather than holding references to the settings object.

I also included the preceding commit (cb87d0f) by way of contrast, which keeps the lifetimes and the references to settings strings, in case that's preferred instead.

philbooth:
Putting the `WIP` back on this, I forgot to test how those futures changes work for real and it turns out `loop_fn` isn't looping like I wanted it to.  I'll fix it tomorrow, but any comments on this are still welcome in the meantime of course.

philbooth:
> ...`loop_fn` isn't looping like I wanted it to.

Nvm, I'd just forgotten to call `wait` on the loop promise.

src/queues/sqs.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+}
+
+impl From<JsonError> for QueueError {
+    fn from(error: JsonError) -> QueueError {
+        QueueError::new(format!("JSON error: {:?}", error))
+    }
+}
+
+enum SqsWrapper {
```
rfk: I've gotta admit, I don't really understand what this is for.  Is there a way to explain it in a comment here, or does it depend too much on the details of the above compiler error you mentioned?

src/queues/mock.rs
```diff
@@ -0,0 +1,136 @@
...
+                            "fxa-email-service.queues.mock.complaint@example.com",
+                        ),
+                    }],
+                    timestamp: Utc::now(),
+                    feedback_id: String::from(""),
+                    user_agent: None,
+                    complaint_feedback_type: None,
+                    arrival_date: Utc::now(),
+                });
+                insane_message
```
rfk: nit: something about the use of "insane" throughout this doesn't sit right with me, can we use a less emotional term like "unexpected"?

src/queues/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+            NotificationType::Delivery => Ok(()),
+            NotificationType::Null => {
+                Err(QueueError::new(String::from("Insane notification type")))
+            }
+        };
+        match result {
+            Ok(_) => {
+                let future = self
+                    .notification
+                    .send(&notification)
```
rfk: So IIUC, we're basically pulling SES notifications from three separate "incoming" queues, recording them in the db, and then forwarding them over to the auth-server via a single "outgoing" queue.  Is that accurate?

I wonder whether it's worth teasing the "incoming" and "outgoing" queues apart a little more, because the incoming queues are and will always be very specific to the details of SES, but ISTM the outgoing queue should be more generic and hide some of those details.  For example, does it make sense to have an outgoing `Notification` type that doesn't include SES-specific details like `mail.source_arn`?

src/queues/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+#[derive(Debug)]
+pub struct QueueError {
+    description: String,
+}
+
+#[derive(Debug)]
+pub struct QueueIds {
+    pub bounce: String,
+    pub complaint: String,
+    pub delivery: String,
```
rfk: Is it actually necessary for us to have three distinct SQS queues, one for each of bounces, complains, and delivery notifications?  Or is that a historical artifact of the way we added these in the auth-server, and they could actually all be delivered via the same queue if we configured things appropriately in AWS?

philbooth: I can't find anything in the docs that says you can't use the same queue for all three notification types, so I presume it would be fine. But @jrgm or @jbuck can probably confirm for us: Was it just our own preference to use separate queues for bounce, complaint and delivery notifications or are there objective reasons why it has to be that way?

src/queues/mod.rs
philbooth: > For example, does it make sense to have an outgoing `Notification` type that doesn't include SES-specific details like `mail.source_arn`?

Yep. This is what I meant in the opening comment about the `Notification` format leaking out. It was a shortcut taken to try and reduce the volume of data-marshalling in this code review, and I was thinking YAGNI (at least for now, anyway). Happy to add it though.

src/queues/sqs.rs
philbooth: Yep, will do.

Just to answer it here too, the obvious fix to the problem of `Debug` not being implemented (the compiler error mentioned above), would be just to do something like this:

```rust
impl Debug for SqsClient<CredentialsProvider, RequestDispatcher> {
    ...
}
```

But if you try to do that, you get this error:

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
   --> src/queues/sqs.rs:217:1
    |
217 | impl Debug for SqsClient<CredentialsProvider, RequestDispatcher> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ impl doesn't use types inside crate
    |
    = note: the impl does not reference any types defined in this crate
    = note: define and implement a trait or new type instead
```

So in order to `Box` the SQS client, I needed a way to implement `Debug` but I wasn't allowed to do it directly. So I wrap it in a new type, proxying the methods that I want to call, and implementing `Debug` on the outer type.

But because there are actually two different concrete specialisations of the abstract `SqsClient` type (which is why I needed to `Box` it in the first place of course), I want to wrap both of those concrete types with the same implementation. Thus the wrapping type is an `enum` and I proxy to the appropriate inner type at runtime.

There may, probably are, better ways to do this I suspect. It was a home-baked solution and I didn't spend a great deal of time on it.

philbooth:
> I had a first attempt at reviewing this, but there's a lot going on here! I think I'm going to need a bit more context on the pieces and how they all fit together. Is it possible to add a thousand-foot overview to the readme or some other docs file?

Yeah, sorry about that. I'll add the docs, for sure, but I also think maybe reviewing it all in one go was too much of an ask. I'll have another stab at breaking it down into smaller logical chunks.

src/queues/sqs.rs
philbooth: ...although interestingly, I finally got round to producing a cut-down example of the problem with a view to opening an issue, and it works okay there. So something else I was doing must have been causing it, will fix.

src/bounces/test.rs
```diff
@@ -240,6 +276,15 @@ impl Db for DbMockError {
...
     }
+
+    fn create_bounce(
+        &self,
+        _address: &str,
+        _bounce_type: BounceType,
+        _bounce_subtype: BounceSubtype,
+    ) -> Result<(), DbError> {
+        Err(DbError::new(String::from("Not implemented")))
+    }
```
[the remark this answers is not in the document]
philbooth: So, turns out traits can provide default method implementations. I have a cunning plan to eliminate this repetition.

src/queues/mod.rs
philbooth: > I wonder whether it's worth teasing the "incoming" and "outgoing" queues apart a little more, because the incoming queues are and will always be very specific to the details of SES, but ISTM the outgoing queue should be more generic and hide some of those details.

Just to add a bit more about this part of your comment, because I don't think my intentions were clear. I was aiming for something that can plug in to the existing auth server code and have it just work without many changes needed over there. Because the auth server already has code for emitting flow events from SQS notifications, I figured the natural first step was to get everything working end-to-end using the same message format. As a secondary consideration it was also nice that it made this PR a bit simpler, because it's one less data mapping to implement, test and review.

I agree it's definitely not what we want to do long-term, but quite like the idea of limiting the amount of short-term churn we'll have when we first come to glue everything together.

In terms of teasing them apart some more right now, would it suffice for `Outgoing` to just take a json value? It would just be the direct serialisation of the existing `Notification` type in this PR, but at least it would then be decoupled for the future change when we do come to send our own, more-generic message format.

src/queues/mod.rs
jbuck: I think our own preference. It's possible to deliver to a single queue for all three types

philbooth: > I think our own preference. It's possible to deliver to a single queue for all three types

Thanks!

@rfk, I'm definitely in favour of simplifying it to a single incoming queue from a conceptual standpoint, I just wonder whether there is a benefit to keeping three queues for now from a pragmatic standpoint? I mean, if we did want to run this alongside/interchangeably with the existing auth server bounce-handling code at first, then wouldn't it make sense to do that while consuming from the exact same queues?

philbooth:
> I'll have another stab at breaking it down into smaller logical chunks.

As you can see, I started doing this today a bit. Not sure if it's helpful or not, feel free to ignore those PRs if you'd rather review everything in one go. I'll write some docs about how the queue stuff fits together tomorrow (and, if do we want to merge any of those separate PRs, rebase this one so the diff is zeroed in on just the queue stuff).

src/queues/mod.rs
rfk: Thanks, that makes sense.  I agree there's definitely value in having something that will plug in directly to the auth-server as-is in order to evolve things forward.  I wonder if we can just send a Notification but without any bits that the auth-server won't read, e.g. an "OutgoingNotification" type that has a subset of the fields of the incoming "Notification" type.

src/queues/mod.rs
rfk: Aye, continuing to use the three separate queues sounds for compatibility with auth-server sounds like the right place to start from, 👍

philbooth:
@rfk, this has now been rebased against those smaller changes I extracted and includes a couple of fixes to address your initial feedback:

* The readme contains a description of how the queue stuff works. Not sure if this is what you had in mind or if you wanted something more in depth? If the latter, I'll probably do it in a different doc because it's maybe out of scope for the readme.

* `sqs::Queue` now returns/accepts a "generic" notification type. I say "generic" because I cheated a bit to try and write less code. The generic type just re-exports the nested types that it uses from the SQS notification type instead of defining it's own. But the calling code doesn't know that, it still accesses those types from the generic module. So conceptually it's like a filter rather than a distinct type I guess.

Fwiw, I'm also hoping to get some feedback from the push team in due course, it feels like we're getting to a point here where that's maybe more valuable than it was at the start.

src/queues/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn process_queue(&'static self, queue: &'static Box<Incoming>) -> QueueFuture {
+        let future = queue
+            .receive()
+            .and_then(move |messages| {
+                let mut futures: Vec<Box<Future<Item = (), Error = QueueError>>> = Vec::new();
+                for message in messages.into_iter() {
+                    if message.notification.notification_type != NotificationType::Null {
+                        let future = self
+                            .handle_notification(&message.notification)
+                            .and_then(move |_| queue.delete(message));
```
rfk: Since we receive messages in batches, I wonder if we should also delete them in batches using [1], or whether that would require us to jump through some ugly hoops in the processing loop here.

[1] https://rusoto.github.io/rusoto/rusoto_sqs/trait.Sqs.html#tymethod.delete_message_batch

philbooth:
> ...there's a lot of rust stuff in here that I kind of have to gloss over - for example, a lot of Box usage that I don't really understand but am assuming is a necessary side-effect of using futures here.

Fwiw that's pretty high on my list of things to call out for them this afternoon, I don't like it either. All of the different combinators return their own specific concrete implementation of the `Future` trait for some reason, hence boxing it and passing the trait everywhere instead of using a generic type like `FutureResult` directly. Maybe I'm supposed to use the `From` trait to convert them instead or something, will ask later.

src/queues/mod.rs
philbooth: Oh, good shout. :+1:

philbooth: (I opted to leave this alone for now, but have opened #54 to so we can come back to it)

PR #31, opened 2018-05-28
feat(db): implement AuthDb::create_bounce

philbooth:
Extracted from #28. Depends on #29.

### DIFF
--- a/src/auth_db/test.rs
+++ b/src/auth_db/test.rs
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
+use std::time::SystemTime;
+
 use serde_json;
 
 use super::*;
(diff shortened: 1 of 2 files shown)


philbooth:
@mozilla/fxa-devs r?

You can just review this one and ignore #29 if you'd prefer, or you can review #29 first and then I'll rebase this one after that's merged. Whichever is preferable.

src/auth_db/test.rs
```diff
@@ -26,25 +142,62 @@ fn get_bounces_invalid_address() {
...
-fn deserialize_invalid_bounce_type() {
-    match serde_json::from_value::<BounceType>(json!(4)) {
-        Ok(_) => assert!(false, "serde_json::from_value should have failed"),
-        Err(error) => assert_eq!(error.description(), "JSON error"),
-    }
-    match serde_json::from_value::<BounceType>(json!(-1)) {
-        Ok(_) => assert!(false, "serde_json::from_value should have failed"),
-        Err(error) => assert_eq!(error.description(), "JSON error"),
+    // The email addresses in this test are pseudo-namespaced so that we don't
+    // accidentally create bounce records that cause other tests to fail.
```
rfk: Because tests run in parallel, or because we don't clean them up after the test completes?

src/auth_db/test.rs
```diff
@@ -26,25 +142,62 @@ fn get_bounces_invalid_address() {
...
+    let bounces_after =
+        db.create_bounce(
+            "fxa-email-service.test.auth-db.foo@example.com",
+            BounceType::Hard,
+            BounceSubtype::General,
+        ).and_then(|_| db.get_bounces("fxa-email-service.test.auth-db.foo@example.com"))
+            .expect("db error");
+    let now = now_as_milliseconds();
+
+    if bounces_before.len() == 20 {
```
rfk: It's not at all clear where this magic number `20` comes from, or why.  I assume it's this:

  https://github.com/mozilla/fxa-auth-db-mysql/blob/dd249e5a2a75eca52f620fb8d2acc3e03e8393fd/lib/db/schema/patch-040-041.sql#L42

But that's along way removed for the reader to understand on sight.

philbooth:
> Would adding a random suffix on the email address help here?

Yes it would, good point. I'll fix it, thanks.

src/auth_db/test.rs
philbooth: Also a good point, I'll add a comment.

PR #32, opened 2018-05-28
chore(queues): split out a separate process for queue-handling

philbooth:
Extracted from #28.

Obviously we don't want to do queue-handling in the same process that manages the send API, but we do want to re-use lots of the same code between the two processes. There's a few different ways to achieve that in Rust and I opted for the most basic: just build two different binaries from the same `Cargo.toml`.

I have read some commentary suggesting that a "better" way to do it is put all of the common logic into a separate lib project and then have the two binary projects use that, but I think this way works just as well and involves less monkeying about with directories and `Cargo.toml`. Plus there's some logic that isn't common to both binaries, and it feels more coherent for that to live alongside the common logic than be siloed off elsewhere.

But of course there's nothing to stop us from pursuing the other approach in the future if this one doesn't work out for some reason.

Two shell scripts, `./r` and `./rq`, are provided to save keystrokes when you want to run the respective binaries. There is no attempt made here to actually implement any of the queue-handling logic, that will be in a separate PR.

@mozilla/fxa-devs r?


src/auth_db/test.rs
```diff
@@ -27,23 +27,23 @@ fn get_bounces_invalid_address() {
 
 #[test]
 fn deserialize_bounce_type() {
-    let bounce_type: BounceType = serde_json::from_value(json!(0)).expect("JSON error");
+    let bounce_type: BounceType = serde_json::from_value(From::from(0)).expect("JSON error");
```
philbooth: `json!` comes from Rocket, which is specific to the service binary so can't be used in the common tests.

README.md
```diff
@@ -164,7 +164,7 @@ or in `config/local.json`:
 ```
 
 Once you have config set,
-you can start the service with `cargo run`.
+you can start the service with `./r`.
```
rfk: TBH I'm pretty "meh" on these tiny command aliases, they give me flashbacks to customs-server and all its tiny two-character JSON keys that are hard to understand without context.  But if you find them useful then I'm OK with it :-)

src/auth_db/test.rs
rfk: Is `From::from` here basically converting to a string, and if so would it be clearer to just write the inline string `"0"`?

README.md
philbooth: I do find them useful fwiw. Partly because `cargo run --bin service` is annoying to type, but more because I expect I'll forget exactly what the invocation is any time I haven't run it for a few days.

philbooth: (to be clear in case it wasn't already, `cargo run` on it's own doesn't work any more)

philbooth: Wait, I may have misunderstood your comment. Are you meh because the script doesn't do much or me because the name is so short? I can give more readable names if it helps?

rfk: The later, I'm specifically meh on the short opaque command name that a new reader probably won't understand when they check out the repo and run `ls`.

philbooth: The scripts now live under a `scripts/` directory with descriptive (hopefully) names. I've retained my opaquely-named helpers in the root as symlinks, because I genuinely do find them helpful.

PR #36, opened 2018-06-01
fix(errors): return JSON payloads for error responses

### LINKED ISSUE #8: Implement custom errors

philbooth:
Fixes #8.

It's not a perfect implementation by any means, because there's no request-specific data being included in the payloads. But at least it means we're actually returning JSON for the common error types now.

I opted to include `errno`, `message` and `data` as fields in the `ApplicationError` struct even though we're not setting them yet, just to stake out their existence in the API.

@mozilla/fxa-devs r?


src/send/test.rs
```diff
@@ -95,7 +107,7 @@ fn without_optional_data() {
 fn missing_to_field() {
     let client = setup();
 
-    let response = client
+    let mut response = client
```
vbudhram: Curious why this was changed to be mutable? The code below doesn't appear to be changing the value?

philbooth: It's because `Response::body` is `&mut self`, it returns a mutable reference to the body.

Ref: https://api.rocket.rs/rocket/response/struct.Response.html#method.body

Issue #10, opened 2018-04-27
Implement logging


philbooth:
Probably needs to be mozlog format so that it plays nice with the rest of our logging infrastructure. Should log errors, success, request summaries with timing info et cetera and so on.

philbooth:
Not necessarily right away, but at some point the abstraction we write for this could be published as a crate, if we thought it might be useful on other projects.

philbooth:
To do: ask the push folks if they've already built some kind of mozloggy thing that we could re-use.

philbooth:
> To do: ask the push folks if they've already built some kind of mozloggy thing that we could re-use.

```
16:41:03 <pb> hey push team, i was wondering whether you guys had maybe implemented a mozlog-friendly logger for rust that we could re-use in fxa?
16:44:10 <•benbangert> pb: we do have one, pjenvey is actually extracting it into its own package this week
16:45:49 <•benbangert> pb: it uses the slog crate for structured logging, so we can write log statements like: https://github.com/mozilla-services/autopush/blob/master/autopush_rs/src/client.rs#L514
16:46:11 <•benbangert> and it appropriately nests them under Fields, etc. for Moz's Logging 2.0 JSON format
16:47:42 <pb> benbangert: thanks, that sounds perfect! is there a repo i can watch yet?
16:48:16 <pb> (or will it be inside autopush?)
16:48:34 <•benbangert> its going to be its own crate
16:49:19 <•benbangert> cause there's some annoying setup code to move in there too: https://github.com/mozilla-services/autopush/blob/master/autopush_rs/src/util/mod.rs#L56
```

philbooth:
```
~/c/fxa-email-service (pb/5) $ cargo search mozlog
    Updating registry `https://github.com/rust-lang/crates.io-index`
slog-mozlog-json = "0.1.0"    # MozLog JSON drain for slog-rs
```

philbooth:
* https://github.com/mozilla-services/slog-mozlog-json
* https://docs.rs/slog-mozlog-json/0.1.0/slog_mozlog_json/

philbooth:
Beware that, until we move to rocket 0.4, there will be naming clashes with rocket's `error` macro when we bring in slog.

vladikoff:
@brizental useful: https://github.com/mozilla-services/megaphone/pull/38/files

brizental:
This issue is not done yet, still need to log settings and queue stuff :)

vladikoff:
[...]

PR #37, opened 2018-06-01
fix(sendgrid): return message id from the sendgrid provider

philbooth:
Marking as `WIP` while I wait to see what comes out of gsquire/sendgrid-rs#23. This gets it working but may be unnecessarily hackish.


philbooth:
The `sendgrid` crate got a new release over the weekend, which exposes the whole response object instead of just the status. This PR now uses that to return the message id from the Sendgrid provider.

@mozilla/fxa-devs r?

src/providers/mod.rs
```diff
@@ -89,16 +89,14 @@ impl<'s> Providers<'s> {
             self.default_provider
         };
 
-        match self.providers.get(resolved_provider_id) {
```
philbooth: I'm replacing these `match` expressions with combinators as I find them. I wasn't aware of the combinator stuff when I started this repo.

philbooth:
(so you should now see something like this in the terminal when you send mail via Sendgrid:

```
~/c/fxa-email-service (pb/sendgrid-id) $ curl \
    -d '{"to":"foo@example.com","subject":"bar","body":{"text":"baz"},"provider":"sendgrid"}' \
    -H 'Content-Type: application/json' \
    http://localhost:8001/send
{"messageId":"sendgrid:5Y5lBlvOR3GBp3yzLA5gNA"}
```

src/send/mod.rs
```diff
@@ -116,15 +116,18 @@ fn handler(email: Email) -> Result<Json<Value>, Failure> {
...
+            email.to.as_ref(),
+            cc.as_ref(),
+            email.subject.as_ref(),
+            email.body.text.as_ref(),
+            html,
+            provider,
+        )
+        .map(|message_id| Json(json!({ "messageId": message_id })))
+        .map_err(|error| {
+            println!("{}", error);
```
shane-tomlinson: Do you want the println here?

philbooth: I do (for now). While I was implementing this, stuff was broken for a while and I realised that nowhere was printing the error message. The most helpful thing I had to go on was `"Internal Server Error"`.

Note that soon we'll hopefully have proper logging in this repo, courtesy of #10. I know `println!` looks like a hack in the meantime (entirely reasonably of course, because it is one), but it's better than nothing.

src/providers/sendgrid.rs
```diff
@@ -82,3 +104,11 @@ impl From<SendgridError> for ProviderError {
         }
     }
 }
+
+impl From<Utf8Error> for ProviderError {
+    fn from(error: Utf8Error) -> ProviderError {
+        ProviderError {
+            description: format!("Failed to decode Sendgrid message id as UTF-8: {:?}", error),
```
rfk: The text of this error message runs the risk of accidentally becoming out of date if we ever try to parse some other utf8 string from the sendgrid API, but use the same `From::from` error handling.  Not really sure how to make that better, just calling it out.

PR #45, opened 2018-06-05
bug: allow RFC5321 compliant email addresses

### LINKED ISSUE #44: Allow RFC5321 compliant email addresses
Looks like the code is doing very primitive email verification. We should at the very least allow '+' in the local part.

patch inbound.

jrconlin:
email address validation is a pain, and you can use a surprising variety
of characters in an address.

(At least we're drawing the line at no quoted characters...)

Closes #44

### DIFF
--- a/src/validate/mod.rs
+++ b/src/validate/mod.rs
@@ -15,7 +15,7 @@ lazy_static! {
         "^https?://[A-Za-z0-9-]+(?:\\.[A-Za-z0-9-]+)*(?::[0-9]+)?/(?:[A-Za-z0-9-]+/)*$"
     ).unwrap();
     static ref EMAIL_ADDRESS_FORMAT: Regex =
-        Regex::new("^[a-z0-9-]+@[a-z0-9-]+(?:\\.[a-z0-9-]+)+$").unwrap();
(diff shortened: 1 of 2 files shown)


philbooth:
Thanks @jrconlin!

src/validate/mod.rs
```diff
@@ -15,7 +15,7 @@ lazy_static! {
         "^https?://[A-Za-z0-9-]+(?:\\.[A-Za-z0-9-]+)*(?::[0-9]+)?/(?:[A-Za-z0-9-]+/)*$"
     ).unwrap();
     static ref EMAIL_ADDRESS_FORMAT: Regex =
-        Regex::new("^[a-z0-9-]+@[a-z0-9-]+(?:\\.[a-z0-9-]+)+$").unwrap();
+        Regex::new("^[a-z0-9\\-!#\\$%\\&`\\*\\+/=\\?\\^{\\|}~]+@[a-z0-9-]+(?:\\.[a-z0-9-]+)+$").unwrap();
```
shane-tomlinson: We should probably port this over to the content-server and auth-server validation too.

PR #55, opened 2018-06-05
fix(validation): relax the sendgrid API key validation regex

philbooth:
I thought this might have been too restrictive and it was. I didn't listen to myself though, did I?

@brizental, this should let your Sendgrid API key pass validation. r?

(and @fxa-devs r?)

### DIFF
--- a/src/validate/mod.rs
+++ b/src/validate/mod.rs
@@ -21,7 +21,7 @@ lazy_static! {
     static ref PROVIDER_FORMAT: Regex = Regex::new("^(?:mock|sendgrid|ses|smtp)$").unwrap();
     static ref SENDER_NAME_FORMAT: Regex =
         Regex::new("^[A-Za-z0-9-]+(?: [A-Za-z0-9-]+)*$").unwrap();
-    static ref SENDGRID_API_KEY_FORMAT: Regex = Regex::new("^[A-Za-z0-9._]{69}$").unwrap();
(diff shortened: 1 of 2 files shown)


philbooth:
> my original comment originated from me trying the API KEY ID when it seemed like nothing was working at all.

Yeah, wasn't a reflection on your comment on that PR fwiw, I just figured it will fail anyway if the key is wrong so at least this way we rule out false validation errors for correct keys (like the one @brizental had, which was related to the hyphen rather than length, but the principle still applies I think).

brizental:
Hey @philbooth , I tested this here and the API key error is gone, but I'm getting a `429 Too Many Requests catcher.` error, don't know if it's related to this or not.

philbooth:
> I'm getting a `429 Too Many Requests` error, don't know if it's related to this or not.

Probably not related to this. That should only happen if there's bounce/complaint records in the auth db for one of the email addresses you're sending to. Not sure if it's possible you triggered some bounces with that email address though, it would require you to either be running the queues binary or to have triggered them through your local auth server instead...

philbooth:
> That should only happen if there's bounce/complaint records in the auth db...

Oh wait, maybe there is some other db error and it's being incorrectly coerced to a bounce error. @brizental, we can debug this tomorrow, but I'm thinking maybe it's not connecting to your fxa-auth-db-server process.

philbooth:
(acutally, maybe I didn't mention that it needs to connect to a running fxa-auth-db-server in the readme, I should fix that)

philbooth:
@brizental, if you start the db server with fxa-local-dev, it might start working...

PR #58, opened 2018-06-05
chore(project): limit futures to < 0.2

philbooth:
Whoops, should have remembered this when reviewing #57, we can't use futures 0.2 yet so this pins it to the 0.1 family. (running `cargo update` then `cargo t` without this change will fail)

@mozilla/fxa-devs r?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,15 +115,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "card-validate"
-version = "2.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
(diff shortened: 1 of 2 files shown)


vladikoff:
> we can't use futures 0.2 yet

Do we need a follow up issue for that? Do we need to migrate some code to make it work?

philbooth:
> Do we need a follow up issue for that? Do we need to migrate some code to make it work?

We have to use the same version as rusoto uses. I don't think we need an issue for it because if/when rusoto updates, the build will fail with `Future` errors so it should be pretty obvious what's up. Although having an issue open for it wouldn't do any harm I guess, feel free to open one if you think it's useful.

vladikoff:
@philbooth the comment is useful, thanks!

PR #59, opened 2018-06-06
fix(bounces): ensure db errors don't get converted to 429 responses

### LINKED ISSUE #56: Are db errors being converted to bounce responses?
Not sure, kind of guessing based on the discussion at https://github.com/mozilla/fxa-email-service/pull/55#issuecomment-394835844. Opening as a reminder to dig into it tomorrow.

philbooth:
Fixes #56.

We were indiscriminately converting all bounce errors to a 429 response. This changes it so we only return the 429 if the error has a bounce record, otherwise we return a 500.

@mozilla/fxa-devs r?

### DIFF
--- a/src/bounces/mod.rs
+++ b/src/bounces/mod.rs
@@ -63,10 +63,14 @@ impl From<DbError> for BounceError {
 }
 
 impl From<BounceError> for Failure {
-    fn from(_error: BounceError) -> Failure {
+    fn from(error: BounceError) -> Failure {
         // Eventually we should be able to do something richer than this,
         // as per https://github.com/SergioBenitez/Rocket/issues/586.
-        Failure(Status::TooManyRequests)
+        Failure(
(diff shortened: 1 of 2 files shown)


src/bounces/mod.rs
```diff
@@ -63,10 +63,14 @@ impl From<DbError> for BounceError {
...
 impl From<BounceError> for Failure {
-    fn from(_error: BounceError) -> Failure {
+    fn from(error: BounceError) -> Failure {
         // Eventually we should be able to do something richer than this,
         // as per https://github.com/SergioBenitez/Rocket/issues/586.
-        Failure(Status::TooManyRequests)
+        Failure(
+            error
+                .bounce
+                .map_or(Status::InternalServerError, |_| Status::TooManyRequests),
```
vladikoff: 👍

PR #60, opened 2018-06-06
chore(project): remove the validator dependency

### LINKED ISSUE #41: Delete the validator crate
We're using the `validator` crate to validate email addresses on the `/send` endpoint in `src/send.mod.rs`. That seems like a flimsy justification for an entire dependency, especially when we also have home-baked (and home-tested) validation code `src/validate/mod.rs`.

For simplicity, I think we should ditch `validator` and use the same validation layer for everything.

philbooth:
Fixes #41.

Since we got the sprauncy new regex in #45, I decided to quickly whip out this dependency right away. It's been annoying me for a while.

@mozilla/fxa-devs r?

### DIFF
--- a/src/send/mod.rs
+++ b/src/send/mod.rs
@@ -2,16 +2,14 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
-use std::{borrow::Cow, collections::HashMap};
-
 use rocket::{
     data::{self, FromData}, http::Status, response::Failure, Data, Outcome, Request,
 };
 use rocket_contrib::{Json, Value};
-use validator::{self, Validate, ValidationError};
 
 use auth_db::DbClient;
 use bounces::Bounces;
+use deserialize;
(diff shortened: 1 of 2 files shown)


src/send/mod.rs
```diff
@@ -43,7 +41,7 @@ struct Email {
 }
 
 impl FromData for Email {
-    type Error = ValidationError;
+    type Error = ();
```
vladikoff: I guess this will be changed once we use the suggested error crates?

src/service_main.rs
```diff
@@ -25,9 +25,6 @@ extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde_json;
-extern crate validator;
-#[macro_use]
-extern crate validator_derive;
```
vladikoff: 👍

src/send/mod.rs
philbooth: Possibly, although I think it's more contingent on a resolution to SergioBenitez/Rocket#596 tbh.

Issue #49, opened 2018-06-05
Inject config rather than statically initialising


philbooth:
Statically initialising config makes it hard to test (as we've already discovered). Recommendation from the push folks is to inject it everywhere we need it, so that tests can properly interrogate what's going on.

philbooth:
(and then we can get rid of that rubbish `CleanEnvironment` hack I did)

philbooth:
(and also stop running the tests in a single thread)

brizental:
I'll work on this :) It's necessary to make the settings logging better.

PR #63, opened 2018-06-07
feat(logging): add mozlog logging to requests

### LINKED ISSUE #10: Implement logging

brizental:
Starts to fix #10 

I adapted the megaphone logging module for our purposes. https://github.com/mozilla-services/megaphone/blob/c7353f7e167a34aa7617cc23ac9489f395d77ff4/src/logging.rs

This is using [Fairings](https://api.rocket.rs/rocket/fairing/trait.Fairing.html) to log every time  a user makes a request. It logs when the request started and when it is done. 

There are still some more `println!`'s to replace, specially in the queues, but I thought it was best to send this in for you guys to check if I'm in the right direction.

Also, I added a new config variable `FXA_EMAIL_JSONLOGGING`, it defaults to `false`. If it's `true` it will log mozlog style. If it's `false` it will use pretty formatting.

r? @philbooth @vladikoff

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,22 @@
 [package]
 name = "fxa-email-service"
 version = "0.1.0"
+[[bin]]
+name = "service"
+path = "src/service_main.rs"
+
+[[bin]]
+name = "queues"
+path = "src/queues_main.rs"
 
 [dependencies]
-chrono = { version = ">=0.4.2", features = [ "serde" ] }
 config = ">=0.8.0"
+failure = ">=0.1.1"
 futures = ">=0.1.21,<0.2"
 hex = ">=0.3.2"
 lazy_static = ">=1.0"
 md5 = ">=0.3.7"
+mozsvc-common = "0.1.0"
 regex = ">=1.0"
 reqwest = ">=0.8.5"
 rocket = ">=0.3.12"
@@ -22,13 +30,13 @@ sendgrid = ">=0.7.0"
 serde = ">=1.0"
 serde_derive = ">=1.0"
 serde_json = ">=1.0"
+slog = "2.2.3"
+slog-async = "2.3.0"
+slog-mozlog-json = "0.1.0"
+slog-term = "2.4.0"
 validator = ">=0.6.3"
 validator_derive = ">=0.6.5"
 
-[[bin]]
-name = "service"
-path = "src/service_main.rs"
-
-[[bin]]
-name = "queues"
-path = "src/queues_main.rs"
+[dependencies.chrono]
+features = ["serde"]
+version = ">=0.4.2"
--- a/config/default.json
+++ b/config/default.json
@@ -19,6 +19,7 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
+  "jsonlogging": "false",
   "provider": "ses",
   "sender": {
     "address": "accounts@firefox.com",
--- a/config/default.json
+++ b/config/default.json
@@ -19,7 +19,7 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
-  "jsonlogging": "false",
+  "jsonlogging": "true",
   "provider": "ses",
   "sender": {
     "address": "accounts@firefox.com",
--- a/src/logging/mod.rs
+++ b/src/logging/mod.rs
@@ -1,0 +1,104 @@
+/// Logging via slog
+///
+/// Provides a RequestLogger with moz log fields per request
+ 
+use std::io;
+use std::ops::Deref;
+
+use failure::{err_msg};
+use rocket::{
+    Request, State,
+};
+use slog::{self, Drain, Record, Serializer, KV};
+use slog_async;
+use slog_mozlog_json::MozLogJson;
+use slog_term;
+
+use app_errors::Result;
+use settings::Settings;
+
+lazy_static! {
+    static ref LOGGER_NAME: String =
+        format!("{}-{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"));
+    static ref MSG_TYPE: String = format!("{}:log", env!("CARGO_PKG_NAME"));
+}
+
+#[derive(Clone)]
+struct MozLogFields {
+    method: &'static str,
+    path: String,
+    remote: Option<String>,
+    agent: Option<String>,
+}
+
+impl MozLogFields {
+    pub fn from_request(request: &Request) -> MozLogFields {
+        let headers = request.headers();
+        MozLogFields {
+            method: request.method().as_str(),
+            path: request.uri().to_string(),
+            agent: headers.get_one("User-Agent").map(&str::to_owned),
+            remote: headers
+                .get_one("X-Forwarded-For")
+                .map(&str::to_owned)
+                .or(request.remote().map(|addr| addr.ip().to_string())),
+        }
+    }
+}
+
+impl KV for MozLogFields {
+    fn serialize(&self, _: &Record, serializer: &mut Serializer) -> slog::Result {
+        if let Some(ref agent) = self.agent {
+            serializer.emit_str("agent", agent)?;
+        }
+        if let Some(ref remote) = self.remote {
+            serializer.emit_str("remoteAddressChain", remote)?;
+        }
+        serializer.emit_str("path", &self.path)?;
+        serializer.emit_str("method", self.method)
+    }
+}
+
+pub struct RequestLogger(slog::Logger);
+
+impl RequestLogger {
+    pub fn with_request(request: &Request) -> Result<RequestLogger> {
+        let logger = request
+            .guard::<State<RequestLogger>>()
+            .success_or(err_msg("Internal error: No managed RequestLogger"))?;
+        Ok(RequestLogger(
+            logger.new(slog_o!(MozLogFields::from_request(request))),
+        ))
+    }
+}
+
+impl Deref for RequestLogger {
+    type Target = slog::Logger;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+pub fn init_logging(settings: &Settings) -> Result<RequestLogger> {
+    let json_logging = match settings.jsonlogging {
+        Some(json_logging) => json_logging,
+        None => Err(err_msg("Invalid FXA_EMAIL_JSONLOGGING"))?,
+    };
+
+    let logger = if json_logging {
+        let drain = MozLogJson::new(io::stdout())
+            .logger_name(LOGGER_NAME.to_owned())
+            .msg_type(MSG_TYPE.to_owned())
+            .build()
+            .fuse();
+        let drain = slog_async::Async::new(drain).build().fuse();
+        slog::Logger::root(drain, slog_o!())
+    } else {
+        let decorator = slog_term::TermDecorator::new().build();
+        let drain = slog_term::FullFormat::new(decorator).build().fuse();
+        let drain = slog_async::Async::new(drain).build().fuse();
+        slog::Logger::root(drain, slog_o!())
+    };
+    Ok(RequestLogger(logger))
+}
--- a/src/logging/mod.rs
+++ b/src/logging/mod.rs
@@ -1,6 +1,6 @@
 /// Logging via slog
 ///
-/// Provides a RequestLogger with moz log fields per request
+/// Provides a MozlogLogger with moz log fields per request
  
 use std::io;
 use std::ops::Deref;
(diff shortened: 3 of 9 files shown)


Cargo.toml
```diff
@@ -1,14 +1,22 @@
 [package]
 name = "fxa-email-service"
 version = "0.1.0"
+[[bin]]
```
vladikoff: i wonder why these moved from the bottom, automated?

Cargo.toml
```diff
@@ -22,13 +30,13 @@ sendgrid = ">=0.7.0"
...
 validator_derive = ">=0.6.5"
 
-[[bin]]
-name = "service"
-path = "src/service_main.rs"
-
-[[bin]]
-name = "queues"
-path = "src/queues_main.rs"
+[dependencies.chrono]
```
vladikoff: custom dependency also changed.  *shrugs*

config/default.json
```diff
@@ -19,6 +19,7 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
+  "jsonlogging": "false",
```
vladikoff: @philbooth is this okay for 'default' and let prod set mozlog or should we set jsonlogging to false in local?

Cargo.toml
```diff
@@ -1,14 +1,22 @@
...
 
 [dependencies]
-chrono = { version = ">=0.4.2", features = [ "serde" ] }
 config = ">=0.8.0"
+failure = ">=0.1.1"
 futures = ">=0.1.21,<0.2"
 hex = ">=0.3.2"
 lazy_static = ">=1.0"
 md5 = ">=0.3.7"
+mozsvc-common = "0.1.0"
```
vladikoff: for deps can we do `>`

src/send/mod.rs
```diff
@@ -3,10 +3,10 @@
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
 use rocket::{
-    data::{self, FromData}, http::Status, response::Failure, Data, Outcome, Request,
```
vladikoff: undo this

src/settings/mod.rs
```diff
@@ -92,11 +92,12 @@ pub struct Settings {
     pub authdb: AuthDb,
     pub aws: Aws,
     pub bouncelimits: BounceLimits,
+    pub jsonlogging: Option<bool>,
     #[serde(deserialize_with = "deserialize::provider")]
     pub provider: String,
     pub sender: Sender,
     pub sendgrid: Option<Sendgrid>,
-    pub smtp: Smtp,
+    pub smtp: Smtp
```
vladikoff: did vscode clean this up?

config/default.json
vladikoff: commit `dev.json` with this `false`, set `true` to prod

src/logging/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+            serializer.emit_str("remoteAddressChain", remote)?;
+        }
+        serializer.emit_str("path", &self.path)?;
+        serializer.emit_str("method", self.method)?;
+
+        Ok(())
+    }
+}
+
+impl Value for Settings {
```
brizental:
```
{"Logger":"fxa-email-service-0.1.0","Type":"fxa-email-service:log","Pid":14047,"Severity":6,"Timestamp":1528418741089779000,"Fields":{"msg":"Finished configuration.","auth_db_uri":"http://127.0.0.1:8000/","provider":"ses","aws_keys_access":"[not set]","aws_region":"us-east-1","aws_sqs_urls":"[not set]","json_logging":"true","sendgrid_key":"[hidden]","smtp_host":"127.0.0.1","smtp_port":"25","smtp_user":"[not set]","smtp_password":"[not set]","bounce_limits_enabled":"true","bounce_limits_complaint":"[BounceLimit { period: 86400000, limit: 0 }, BounceLimit { period: 31536000000, limit: 1 }]","bounce_limits_hard":"[BounceLimit { period: 86400000, limit: 0 }, BounceLimit { period: 31536000000, limit: 1 }]","bounce_limits_soft":"[BounceLimit { period: 300000, limit: 0 }]"}}
```

Cargo.toml
```diff
@@ -22,13 +30,13 @@ sendgrid = ">=0.7.0"
 serde = ">=1.0"
 serde_derive = ">=1.0"
 serde_json = ">=1.0"
+slog = "2.2.3"
+slog-async = "2.3.0"
+slog-mozlog-json = "0.1.0"
+slog-term = "2.4.0"
```
philbooth: `>=` here too.

config/default.json
```diff
@@ -19,6 +19,7 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
+  "jsonlogging": "true",
```
philbooth: Bike-shedding here, but let's call this `mozlog` so it specifically identifies the output format that all our data processing pipelines rely on.

Cargo.toml
philbooth: Fwiw I *kind* of prefer the dependencies all in one big list like they were before. Also it keeps this diff focused on the substantive change, which is supporting mozlog.

src/logging/mod.rs
```diff
@@ -0,0 +1,162 @@
+/// Logging via slog
```
philbooth: Can you add the MPL preamble comment to the top of this module? Just copy it from one of the others...

src/send/mod.rs
```diff
@@ -3,10 +3,9 @@
...
+    data::{self, FromData}, http::Status, response::Failure, Data, Outcome, Request
 };
 use rocket_contrib::{Json, Value};
-
```
philbooth: By convention we group `use` statements into `std`, external and internal, with a blank line separating each group. So this blank line is significant, let's put it back.

src/service_main.rs
```diff
@@ -25,19 +25,31 @@ extern crate serde;
...
 extern crate serde_json;
+extern crate failure;
+#[macro_use(
+   slog_o, slog_info, slog_kv, slog_log, slog_record, slog_b, slog_record_static, slog_error
+)]
+extern crate slog;
+extern crate slog_async;
+extern crate slog_mozlog_json;
+extern crate slog_term;
+extern crate mozsvc_common;
```
philbooth: Let's keep this list in alphebetical order so it's easier to manage.

src/settings/mod.rs
```diff
@@ -5,6 +5,7 @@
 use std::env;
 
 use config::{Config, ConfigError, Environment, File};
+use logging;
```
philbooth: This `use` should move down to underneath `use deserialize`.

src/logging/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+
+impl Deref for MozlogLogger {
+    type Target = slog::Logger;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+pub fn init_logging(settings: &Settings) -> Result<MozlogLogger> {
```
philbooth: Can we make this a member of `MozlogLogger` and rename it to just `init`? I'm thinking it might be nicer to invoke it as `MozLogger::init` (or maybe `MozLogger::new`), since that associates it more strongly with the type its returning.

src/settings/mod.rs
```diff
@@ -131,8 +133,8 @@ impl Settings {
 
         match config.try_into::<Settings>() {
             Ok(settings) => {
-                // TODO: replace this with proper logging when we have it
-                println!("config: {:?}", settings);
+                let logger = logging::init_logging(&settings).unwrap();
```
philbooth: Better than an `unwrap()` is an `expect(...)` that prints some useful error info.

But even better than that, maybe, is propagating the error to the caller, seeing as this is inside a function that already returns errors. Let's try and minimise the number of places that can panic to as few as possible. And maybe one day we'll want to do more thorough unit-testing of the error paths, which can't happen if the code panics on the inside.

src/service_main.rs
```diff
@@ -25,19 +25,31 @@ extern crate serde;
...
 mod deserialize;
 mod duration;
+mod logging;
 mod providers;
 mod send;
 mod settings;
 mod validate;
 
 fn main() {
+    let logger = logging::init_logging(&settings::Settings::new().unwrap()).unwrap();
```
philbooth: Hmm, this smells a bit funny because we're instantiating `Settings` twice in this process now. Once here, and once in `src/send/mod.rs`. That is mostly my fault for using `lazy_static` rather than a sensible injection strategy (see #49).

What I think would be better is if we `manage(settings)` below like you already are for `logger`. Then the `send` module could just add it as a param, see the Rocket docs about managed state for more on this:

https://rocket.rs/guide/state/#managed-state

src/logging/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+
+impl Value for Settings {
+    fn serialize(&self, _: &Record, _: Key, serializer: &mut Serializer) -> slog::Result {
+        serializer.emit_str("auth_db_uri", &self.authdb.baseuri)?;
+        serializer.emit_str("provider", &self.provider)?;
+
+        if let Some(_) = self.aws.keys {
+            serializer.emit_str("aws_keys_access", "[hidden]")?;
+        } else {
+            serializer.emit_str("aws_keys_access", "[not set]")?;
```
philbooth: This is good, should we also include `keys.secret`?

src/logging/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        if let Some(_) = self.smtp.password {
+            serializer.emit_str("smtp_password", "[hidden]")?;
+        } else {
+            serializer.emit_str("smtp_password", "[not set]")?;
+        }
+
+        serializer.emit_str("bounce_limits_enabled",  &format!("{}", &self.bouncelimits.enabled))?;
+        serializer.emit_str("bounce_limits_complaint",  &format!("{:?}", &self.bouncelimits.complaint))?;
+        serializer.emit_str("bounce_limits_hard",  &format!("{:?}", &self.bouncelimits.hard))?;
+        serializer.emit_str("bounce_limits_soft",  &format!("{:?}", &self.bouncelimits.soft))?;
```
philbooth: I really like this function but it's pretty verbose. This will do for now but longer-term we should think about how we might be able to eradicate some of the repeated boilerplate code here.

Like would it be possible to have two static lists that define the sensitive and non-sensitive properties and just iterate through the lists? Or would it be possible to annotate them with a macro in `src/settings/mod.rs` so that we can determine whether or not they're sensitive programmatically?

Anything that makes it less likely for us to accidentally leak sensitive data to the logs is worth thinking about.

src/logging/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        }
+        serializer.emit_str("path", &self.path)?;
+        serializer.emit_str("method", self.method)?;
+
+        Ok(())
+    }
+}
+
+impl Value for Settings {
+    fn serialize(&self, _: &Record, _: Key, serializer: &mut Serializer) -> slog::Result {
```
philbooth: As this function is pretty important, we need to come up with a good way to test it. But I'm not sure how tricky it will be in practice to stand up an appropriate `Serializer` argument.

If that proves to be easy then great, if not maybe we can pull out the inner functionality to something more testable. That sounds quite hand-wavey I realise, so maybe we can hack on it in San Francisco if it's not straightforward.

philbooth:
> ...but I thought it was best to send this in for you guys to check if I'm in the right direction.

Looking pretty good I think, definitely moving in the right direction! 👍

vladikoff:
@philbooth @brizental Let's see if we can split this PR a bit, for example, can we just land the new dependencies plus the request logger as `mozlog` first. Then worry about the settings logger and the rest.

src/settings/mod.rs
brizental: idk

brizental: undo this

src/logging/mod.rs
brizental: Since we are not going to show either keys nor access, i thought it was best to just ground them together and show `[hidden]` or `[not set]`. I guess this got a little bit confusing, because I called the property `aws_keys_access`. I'll just change it to `aws_keys`. Sounds better?

src/logging/mod.rs
brizental: I agree. I'll open an issue for that after this PR goes through :)

src/service_main.rs
brizental: Phil, I realized this logger line is not needed for now. That was leftover code from some experiments we were doing earlier.

Anyways, managed stated would be nice. I can work on that after I'm finished with this :)

brizental: I agree with that. I think probably it would be best to send this in as another PR though, fixing specifically the #49 bug. I'll work on that.

brizental:
Closing this because I will send in some smallers PR's with the same content. e.g. #65

Issue #24, opened 2018-05-17
Set rocket config for stage and prod


philbooth:
There is a config file for Rocket in the root directory, `Rocket.toml`. Most of the stuff in there is pretty standard, address and port to listen on, etc. You can read more about it at:

https://api.rocket.rs/rocket/config/

One that stands out as interesting for prod is `workers` / `ROCKET_WORKERS`, which defines how many worker threads Rocket should create for handling requests. The default value is CPU count * 2. We'll probably need to do some testing to zero in on the optimal value for that, especially given that it will be sharing a box with auth server and auth db processes.

vladikoff:
@brizental Could you take a look at this for this train.

I think we need to make sure to tell DevOps to run the app with

ROCKET_ENV=production 

(plus stage support as well)


and have the following configs added based on the doc link above:

address = "127.0.0.1"
port = "8001"

Probably the same for both prod and stage. I wouldn't worry about the `workers` settings for now.

PR #64, opened 2018-06-08
feat(ci): add Dockerhub deploy via CircleCI

### LINKED ISSUE #47: Add the circleci setup based on pushbox
Ref: https://github.com/mozilla-services/pushbox/tree/master/.circleci

vladikoff:
Fixes #47

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,90 @@
+# These environment variables must be set in CircleCI UI
+#
+# DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
+# DOCKER_EMAIL   - login info for docker hub
+# DOCKER_USER
+# DOCKER_PASS
+#
+version: 2
+jobs:
+  build:
+    docker:
+      - image: docker:18.03.0-ce
+    working_directory: /dockerflow
+    steps:
+      - run:
+          name: Install Docker build dependencies
+          command: apk add --no-cache openssh-client git
+      - checkout
+      - setup_remote_docker
+
+      - run:
+          name: Create a version.json
+          command: |
+            # create a version.json per https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md
+            printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' \
+            "$CIRCLE_SHA1" \
+            "$CIRCLE_TAG" \
+            "$CIRCLE_PROJECT_USERNAME" \
+            "$CIRCLE_PROJECT_REPONAME" \
+            "$CIRCLE_BUILD_URL" > version.json
+
+      - run:
+          name: Build Docker image
+          command: docker build -t app:build .
+
+      # save the built docker container into CircleCI's cache. This is
+      # required since Workflows do not have the same remote docker instance.
+      - run:
+          name: docker save app:build
+          command: mkdir -p /cache; docker save -o /cache/docker.tar "app:build"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,90 @@
...
+      - run:
+          name: Restore Docker image cache
+          command: docker load -i /cache/docker.tar
+
+      - run:
+          name: Deploy to Dockerhub
+          command: |
+            # deploy master
+            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+              docker login -u $DOCKER_USER -p $DOCKER_PASS
```
vladikoff: @jbuck could you please add the required creds to make this work into https://circleci.com/gh/mozilla/fxa-email-service/edit#env-vars

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+COPY --from=builder /app/bin /app/bin
+
+WORKDIR /app
+USER app
+
+# override rocket's dev env defaulting to localhost
+ENV ROCKET_ADDRESS 0.0.0.0
+
+CMD ["/app/bin/service"]
```
vladikoff: This is interesting, we need 2 processes here

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    apt-get -qq install -y default-libmysqlclient-dev libssl-dev && \
+    rm -rf /var/lib/apt/lists
+
+COPY --from=builder /app/bin /app/bin
+
+WORKDIR /app
+USER app
+
+# override rocket's dev env defaulting to localhost
+ENV ROCKET_ADDRESS 0.0.0.0
```
vladikoff: the 2 process should have different ports as well

Dockerfile
philbooth: Except we probably don't want the `queues` process running automatically on every box, it should be configurable. Not sure how straightforward that is, but if it's a problem maybe we just have a separate `Dockerfile` for a box that will be spun up to process the queues.

Dockerfile
philbooth: This can also be set in `Rocket.toml` fwiw. We already set the port in there, not sure which is preferable from an ops standpoint. (and we have #24 open for other settings that can go in there, like how many threads to run on)

philbooth: https://github.com/mozilla/fxa-email-service/blob/master/Rocket.toml

philbooth: (it can have separate sections for dev, stage and prod or whatever)

philbooth:
@vladikoff, do you mind if we delete the Circle webhook until this is merged? It's pretty annoying that every single PR gets a big red X next to it for no reason in the meantime...

Dockerfile
pjenvey: it's probably easier for ops to configure anything via env variables so they don't have to configure a volume mount for a config file. of course, ROCKET_PORT is already supported as an env var

Dockerfile
vladikoff: >  we probably don't want the queues process running automatically on every box

We had trouble figuring this out in the past where only certain boxes should have certain processes. I'm gonna add this to dev ops meeting notes to chat about

vladikoff:
close in favour of https://github.com/mozilla/fxa-email-service/pull/87

.circleci/config.yml
jbuck: Done

vladikoff: Thanks!!!

PR #65, opened 2018-06-08
fix(logging): add mozlog logging for requests

### LINKED ISSUE #10: Implement logging

brizental:
As discussed in #63 , this is a partial PR to fix #10 .

I adapted the megaphone logging module for our purposes. https://github.com/mozilla-services/megaphone/blob/c7353f7e167a34aa7617cc23ac9489f395d77ff4/src/logging.rs

This is using [Fairings](https://api.rocket.rs/rocket/fairing/trait.Fairing.html) to log every time a user makes a request. It logs when the request started and when it is done and if it errored.

It uses a config variable called `FXA_EMAIL_MOZLOG` which defaults to `true`. I also added a `dev.json` and, if you set your `NODE_ENV` to `dev` it will be used and will set `FXA_EMAIL_MOZLOG` to `false`, which will produce a prettier log.

I think this is ready to be integrated  with our codebase. I will send other PR's with the settings and the queue loggers.

r? @philbooth @vladikoff


src/logging/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+impl Deref for MozlogLogger {
+    type Target = slog::Logger;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
vladikoff: line here

src/service_main.rs
```diff
@@ -25,19 +25,33 @@ extern crate serde;
...
 extern crate serde_json;
+extern crate failure;
+extern crate mozsvc_common;
+extern crate slog_async;
+extern crate slog_mozlog_json;
+extern crate slog_term;
+#[macro_use(
+   slog_o, slog_info, slog_kv, slog_log, slog_record, slog_b, slog_record_static, slog_error
+)]
+extern crate slog;
```
vladikoff: strange sort order here, the main slog got put all the way to the bottom. *Shrugs*

src/service_main.rs
```diff
@@ -25,19 +25,33 @@ extern crate serde;
...
 mod send;
 mod settings;
 mod validate;
 
 fn main() {
+    let config = settings::Settings::new().expect("Config error.");
+    let logger = logging::MozlogLogger::new(&config).unwrap(); 
     rocket::ignite()
+        .manage(config)
+        .manage(logger)
```
vladikoff: nice!

src/service_main.rs
```diff
@@ -25,19 +25,33 @@ extern crate serde;
...
 mod duration;
+mod logging;
 mod providers;
 mod send;
 mod settings;
 mod validate;
 
 fn main() {
+    let config = settings::Settings::new().expect("Config error.");
+    let logger = logging::MozlogLogger::new(&config).unwrap();
```
vladikoff: can we use `expect`  here? or is this different

src/logging/mod.rs
brizental: done.

src/app_errors/mod.rs
```diff
@@ -2,13 +2,16 @@
...
-use std::collections::HashMap;
+use std::{collections::HashMap, result};
 
+use failure::Error;
 use rocket_contrib::Json;
 
 #[cfg(test)]
 mod test;
 
+pub type Result<T> = result::Result<T, Error>;
```
philbooth: May be worth calling this something like `FxaResult`, then we can use `Result` unnamespaced and we also rule out any chance of future confusion from someone mistaking our result type for the standard one. (and also it's better greppability I guess)

philbooth: Or maybe `AppResult` as this module is `app_errors`. Or something else...

src/logging/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use failure::err_msg;
+use rocket::{Request, State};
+use slog::{self, Drain, Record, Serializer, KV};
+use slog_async;
+use slog_mozlog_json::MozLogJson;
+use slog_term;
+
+use app_errors::Result;
+use settings::*;
```
philbooth: Fwiw, I think I prefer importing specific types from each module explicitly, rather than everything in one go. It just makes it a bit more obvious to see where imports are coming from when you're reading the code.

src/service_main.rs
```diff
@@ -24,20 +24,33 @@ extern crate sendgrid;
 extern crate serde;
 #[macro_use]
 extern crate serde_derive;
+extern crate failure;
+extern crate mozsvc_common;
```
philbooth: These two still don't look alphabetical?

src/service_main.rs
```diff
@@ -24,20 +24,33 @@ extern crate sendgrid;
...
 mod deserialize;
 mod duration;
+mod logging;
 mod providers;
 mod send;
 mod settings;
 mod validate;
 
 fn main() {
+    let config = settings::Settings::new().expect("Config error.");
```
philbooth: Seems wrong that we're doing this before the inject-settings change, but I guess it's fine if we know that one's coming soon? :man_shrugging:

src/app_errors/mod.rs
vladikoff: This was based on megaphone https://github.com/mozilla-services/megaphone/blob/b44f47bc11f51fe6cb4f54479cb202cec64a554c/src/error.rs#L20 but yea, this does look a bit sketchy to call this `Result`.

src/logging/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+
+pub struct MozlogLogger(slog::Logger);
+
+impl MozlogLogger {
+    pub fn new(settings: &Settings) -> Result<MozlogLogger> {
+        let json_logging = match settings.mozlog {
+            Some(json_logging) => json_logging,
+            None => Err(err_msg("Invalid FXA_EMAIL_MOZLOG"))?,
```
philbooth: Rather than do this, let's just remove the `Option` part from the setting so that we can guarantee it's always there. Basically stuff that is already in `config/default.json`, there's no need for it to be optional because of the way options stack.

philbooth:
Thanks @brizental, left some nits but this is close!

src/service_main.rs
brizental: Yeah, I only realized that after the PR was already sent. I guess we can also wait for that one before merging this one. When the inject-settings is accepted I rebase this one, etc.

brizental:
Fixed all the nits and rebased this. Let me know what you think, @philbooth :)

vladikoff:
Great work @brizental and @philbooth!

PR #66, opened 2018-06-09
fix(config): inject config in rocket managed state

### LINKED ISSUE #49: Inject config rather than statically initialising

brizental:
Fixes #49 

Send test are failing at the moment, but ready for review.

r? @philbooth @vladikoff 

Co-authored-by: Philip Jenvey <pjenvey@underboss.org>

### DIFF
--- a/src/auth_db/mod.rs
+++ b/src/auth_db/mod.rs
@@ -3,7 +3,8 @@
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
 use std::{
-    error::Error, fmt::{self, Display, Formatter},
+    error::Error,
+    fmt::{self, Display, Formatter},
 };
 
 use hex;
@@ -273,7 +274,3 @@ impl Db for DbClient {
         }
     }
 }
-
-unsafe impl Sync for DbClient {}
-
-unsafe impl Send for DbClient {}
--- a/src/auth_db/mod.rs
+++ b/src/auth_db/mod.rs
@@ -211,7 +211,7 @@ impl DbUrls {
     }
 }
 
-pub trait Db {
+pub trait Db: Sync {
     fn get_bounces(&self, address: &str) -> Result<Vec<BounceRecord>, DbError>;
 
     fn create_bounce(
@@ -275,3 +275,5 @@ impl Db for DbClient {
 }
 
 unsafe impl Sync for DbClient {}
+
+unsafe impl Send for DbClient {}
--- a/src/bounces/mod.rs
+++ b/src/bounces/mod.rs
@@ -3,7 +3,10 @@
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
 use std::{
-    collections::HashMap, error::Error, fmt::{self, Display, Formatter}, time::SystemTime,
+    collections::HashMap,
+    error::Error,
+    fmt::{self, Display, Formatter},
+    time::SystemTime,
 };
 
 use rocket::{http::Status, response::Failure};
@@ -118,11 +121,7 @@ where
     }
 }
 
-unsafe impl<D> Sync for Bounces<D>
-where
-    D: Db,
-{
-}
+unsafe impl<D> Sync for Bounces<D> where D: Db {}
 
 fn is_bounce_violation(count: u8, created_at: u64, now: u64, limits: &Vec<BounceLimit>) -> bool {
     for limit in limits.iter() {
--- a/src/bounces/mod.rs
+++ b/src/bounces/mod.rs
@@ -74,16 +74,19 @@ impl From<BounceError> for Failure {
     }
 }
 
-pub struct Bounces<'a> {
-    db: Box<&'a Db>,
-    limits: &'a BounceLimits,
+pub struct Bounces<D: Db> {
+    db: D,
+    limits: BounceLimits,
 }
 
-impl<'a> Bounces<'a> {
-    pub fn new(settings: &'a Settings, db: Box<&'a Db>) -> Bounces<'a> {
+impl<D> Bounces<D>
+where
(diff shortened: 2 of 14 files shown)


src/bounces/mod.rs
```diff
@@ -74,21 +74,18 @@ impl From<BounceError> for Failure {
...
         Bounces {
-            db,
-            limits: &settings.bouncelimits,
+            limits: settings.bouncelimits.clone(),
         }
     }
 
-    pub fn check(&self, address: &str) -> Result<(), BounceError> {
-        let bounces = self.db.get_bounces(address)?;
+    pub fn check(&self, address: &str, bounces: Vec<BounceRecord>) -> Result<(), BounceError> {
```
philbooth: So two things here I guess:

1. Now that it's coming in from the outside, let's call the argument `bounce_records` to be a bit more explicit about it and improve greppability.

2. I find the new API for this type a little bit clunky, making the callers do extra legwork in calling the db method on our behalf then passing in the result. Rather than putting `Db` into managed state, can we make it so that only `Bounces` is in managed state and the `Db` instance is retained as a member of `Bounces`? We'd need to put back a bit of the lifetime stuff to make that work, but I think it would be a friendlier interface.

philbooth: I realise, in writing that, my points contradict each other. If we do `2`, which is preferable I think, then we can ignore all the stuff I said about param names here and elsewhere.

src/service_main.rs
```diff
@@ -37,7 +37,16 @@ mod settings;
 mod validate;
 
 fn main() {
+    let config = settings::Settings::new().expect("Config error.");
```
philbooth: Minor nit but can we call this `settings` so it matches the type name?

src/send/mod.rs
```diff
@@ -3,27 +3,21 @@
...
-use auth_db::DbClient;
-use bounces::Bounces;
 use deserialize;
 use providers::Providers;
-use settings::Settings;
 use validate;
 
+use auth_db::{Db, DbClient};
+use bounces::Bounces;
+use settings::Settings;
```
philbooth: These should still be part of the above block I think?

src/bounces/mod.rs
vladikoff: > can we make it so that only Bounces is in managed state and the Db instance is retained as a member of Bounces? We'd need to put back a bit of the lifetime stuff to make that work, but I think it would be a friendlier interface.

This fact that managed state can only deal with `'static` stuff it makes it really complex to deal with db inside of Bounces.  You can experience the problems with that by running this branch: https://github.com/brizental/fxa-email-service/blob/brizental/inject-config-not-lazy/src/service_main.rs#L42

After chatting @pjenvey it was advised to us to separate the DB and Bounces

> [17:16:23] <pjenvey>		I think that's the easiest way to solve that one	
[17:18:27] <pjenvey>		Providers is a little more involved, what needs to happen w/ it is since rocket mandates all these managed pieces to be 'static, what we really need is all the Providers to own their values they pull from config	
[17:18:56] <pjenvey>		which means a few clones() and the like here and there, which is nice to avoid but really doesn't matter here because it's done one time at startup	
[17:20:30] <pjenvey>		that means you won't be able to toggle changes in your Settings during runtime.. but your that should probably be the case anyway	
[17:20:40] <pjenvey>		(change them and have them propagate)

vladikoff: Missed the earlier part from IRC:

> [17:07:55] <pjenvey>		do any of your handlers make raw db calls? solving bounces would be pretty straightforward if it didn't make the db call here	
[17:08:05] <pjenvey>		say the handler made the db call itself, then called bounces

philbooth: > This fact that managed state can only deal with `'static` stuff it makes it really complex to deal with db inside of Bounces.

I think that can be fixed if `Bounces` is generic, so it has a concrete `Db` implementation instead of `Box<Db>`. Then there's no need for any lifetimes and it should just work, right?

brizental:
I finished folding all the changes from #67 into this PR and also fixing the stuff you mentioned in the review @philbooth :) r?

src/auth_db/mod.rs
```diff
@@ -275,3 +275,5 @@ impl Db for DbClient {
 }
 
 unsafe impl Sync for DbClient {}
+
+unsafe impl Send for DbClient {}
```
philbooth: I think we can delete all the `Send` and `Sync` stuff from this module, now that it's not managed by Rocket. Not that it's doing any harm of course, but if we don't need it we don't need it.

brizental: Alright :) 

Also, I just saw that the checks didn't pass. That's because there's a new version of rust and `cargo fmt -- --check` has been updated. I'll fix that as well.

src/providers/mod.rs
```diff
@@ -69,7 +69,7 @@ impl<'s> Providers<'s> {
         }
 
         Providers {
-            default_provider: &settings.provider,
+            default_provider: settings.provider.to_owned(),
```
philbooth: Not that it makes much/any difference, but `to_owned` feels mildly suprising as a reader given that `settings.provider` is already a `String` rather than a `&str`. Would `.cloned()` be clearer?

philbooth:
@brizental this looks great! I left two more tiny nits, sorry I'm such a fusspot!

To get the build passing, I think you need to do a `rustup update` before running `cargo fmt`. Looks like your code is formatted with a different version is all, hence the diffs showing up in the build log. Otherwise, perfect! 👌

PR #67, opened 2018-06-09
refactor(service): move bounces and providers to rocket managed state

philbooth:
Related to #49 and https://github.com/mozilla/fxa-email-service/pull/66#discussion_r194215830. @brizental and @vladikoff this is what I meant by making `Bounces` generic, not sure if I was very clear in my comment. It lets us keep the db stuff out of the calling module is all. Feel free to ignore if you think it's stupid!

### DIFF
--- a/src/send/test.rs
+++ b/src/send/test.rs
@@ -3,14 +3,26 @@
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
 use rocket::{
-    self, http::{ContentType, Status}, local::Client,
+    self,
+    http::{ContentType, Status},
+    local::Client,
 };
 use serde_json;
 
 use app_errors::{self, ApplicationError};
+use auth_db::DbClient;
+use bounces::Bounces;
+use providers::Providers;
+use settings::Settings;
 
 fn setup() -> Client {
+    let settings = Settings::new().unwrap();
+    let db = DbClient::new(&settings);
+    let bounces = Bounces::new(&settings, db);
+    let providers = Providers::new(&settings);
     let server = rocket::ignite()
+        .manage(bounces)
+        .manage(providers)
         .mount("/", routes![super::handler])
         .catch(errors![
             app_errors::bad_request,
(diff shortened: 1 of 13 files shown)


vladikoff:
This looks pretty good at first glance! I am happy this gets rid of the lifetimes.

src/send/test.rs
```diff
@@ -8,9 +8,19 @@ use rocket::{
...
+use auth_db::DbClient;
+use bounces::Bounces;
+use providers::Providers;
+use settings::Settings;
 
 fn setup() -> Client {
+    let settings = Settings::new().unwrap();
+    let db = DbClient::new(&settings);
+    let bounces = Bounces::new(&settings, db);
+    let providers = Providers::new(&settings);
```
vladikoff: In wonder if `providers` or some othe struct need `db` in the future, would this work or not:

```Rust
let bounces = Bounces::new(&settings, db);
let providers = Providers::new(&settings, db);
let thins = Things::new(&settings, db);
```

**shrugs**

philbooth: Fwiw, if `providers` needs access to `auth_db` it means something has gone seriously wrong in our separation of concerns. :smile:

The only thing this repo should have to do with the auth db is reading bounce records and writing bounce records, and even those are temporary until we migrate to a separate data store. Restricting access to where it's needed is a feature, imho.

philbooth: (in that sense, the `bounces` module represents our bespoke interface for the db, we want to hide everything else behind it)

vladikoff: (and writing this makes me realise the queues process should also be accessing it via bounces too now) 

😄

philbooth:
> Also, I saw that you put just the bounces and providers in the managed state, which is fine for now, but we will need the settings in there for the logging stuff.

Yeah, just to be clear, it's not my intention to replace that other PR with this one. I just wanted to show a working/tests-passing solution that keeps the db module inside `bounces`. If you want to fold these change into your own PR that's cool (or if we leave them as two separate PRs that's also fine). The main thing was just to make sure I wasn't talking rubbish when I said we could push `auth_db` into `bounces`.

brizental:
@philbooth I'll just fold these changes into my own PR then :) I think it's easier, probably, to have only one.

PR #71, opened 2018-06-11
feat(clippy): add clippy linting to ci and fix lints

### LINKED ISSUE #52: Use clippy for linting
But probably not in CI because it's a bit of a fuss-pot.

brizental:
Fixes #52 .

For running this locally, just keep in mind that clippy only shows warnings once after each build. That appears to have changed on the updated version of clippy, but since we locked our nightly version we're still having that issue.

Also I added two macros to avoid clippy linting for some stuff. That was because I didn't agree with the lint and wanted to hide the warning message. I tried to add this kinds of configs to the `.clippy.toml` file, but couldn't figure out how, I'm not even sure it's possible.

I added clippy to the end of our CI, also.

Anyways, mostly I think the lints were useful. What do you guys think?

r? @philbooth @vladikoff


src/providers/ses.rs
```diff
@@ -50,7 +50,7 @@ impl Provider for SesProvider {
     ) -> Result<String, ProviderError> {
         let mut destination = Destination::default();
         destination.to_addresses = Some(vec![to.to_string()]);
-        if cc.len() > 0 {
+        if !cc.is_empty() {
```
philbooth: ...like this one for instance. Is it objectively clearer? I'm not sure. Normally I prefer to avoid `!` conditions because they're so easy to misread when you're scanning code quickly.

src/send/mod.rs
```diff
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
+#![cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
```
philbooth: I'm wary of module-level allows in case they permit future us to do something silly. Which code necessitates this one and is there a way to target it more locally to that specific line or whatever?

src/providers/ses.rs
philbooth: (in the other FxA repos we mitigate that to some extent by adding whitespace after the `!`, but unfortunately rustfmt complains when we try to do that here)

src/send/mod.rs
brizental: We can limit this only to the function that we want to bypass. I'll change that.

brizental:
I was thinking now that maybe it's too much of a hassle to formally add clippy to our workflow. 

I think we should keep on using it locally from time to time or something, but adding these `allow` rules everywhere in the code to suppress warnings doesn't really help.

Also, we should merge in some of the code changes that we have here in this PR, but remove clippy from CI and also remove all `allow` rules.

vladikoff:
> Also, we should merge in some of the code changes that we have here in this PR, but remove clippy from CI and also remove all allow rules.

👍

src/queues/mod.rs
```diff
@@ -98,7 +98,7 @@ impl Queues {
             .receive()
             .and_then(move |messages| {
                 let mut futures: Vec<Box<Future<Item = (), Error = QueueError>>> = Vec::new();
-                for message in messages.into_iter() {
+                for message in messages {
```
philbooth: No need to change it, but this one I have mixed feelings about too fwiw. I quite like explicit use of `into_iter()` as a clear signal that the code is consuming `messages` and it can't be used again after the loop.

PR #72, opened 2018-06-12
feat(metadata): store caller-specific metadata with the message id

philbooth:
Related to #4.

Stores a "metadata" string from the caller in Redis, keyed by a hash of the message id. In practice said metadata will be JSON in the auth server's case, but this repo neither knows or cares about that. No functionality is added here to read or clear the data from Redis, that's coming in a separate changeset for the queues process.

No rush to review this, it can wait until you guys are back from the All Hands.


.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: rust
+services:
+  - redis-server
```
vladikoff: 👍

README.md
```diff
@@ -90,6 +90,23 @@ rustup install nightly
...
+```json
+{
+  "redis": {
+    "port": "..."
+  }
+}
+```
+
+You can also set `host` in the same way,
+if your Redis instance is not running locally.
```
vladikoff: 👍

config/default.json
```diff
@@ -19,8 +19,13 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
+  "hmac_key": "YOU MUST CHANGE ME",
```
vladikoff: does this need to have a more verbose name? what if we have keys for other stuff in the future?

src/bin/service.rs
```diff
@@ -11,21 +11,23 @@ extern crate rocket;
...
     let bounces = Bounces::new(&settings, db);
     let logger = MozlogLogger::new(&settings).expect("MozlogLogger::init error");
+    let message_data = MessageData::new(&settings);
     let providers = Providers::new(&settings);
 
     rocket::ignite()
         .manage(bounces)
-        .manage(providers)
         .manage(logger)
+        .manage(message_data)
```
vladikoff: 👍

src/validate/mod.rs
```diff
@@ -45,11 +46,8 @@ pub fn email_address(value: &str) -> bool {
     value.len() < 254 && EMAIL_ADDRESS_FORMAT.is_match(value)
 }
 
-#[cfg(test)]
 pub fn host(value: &str) -> bool {
```
vladikoff: woo, this is actually used now 👍

config/default.json
philbooth: Fwiw, that's why I went for a generic name. If we need to HMAC other data in this repo, can't it just be hashed with the same key?

As long as the key used in prod remains secret, I don't see any benefit to having different keys for various data that needs hashing. And using one key just means there's a little bit less config to manage, which seems like a (small) win.

PR #73, opened 2018-06-13
feat(metadata): send message metadata in outgoing notifications

### LINKED ISSUE #4: Store extra data associated with a request and propagate it on outgoing notifications
Synchronised in the sense that duplicate ids must never occur across peer instances. Should be returned in `200` responses from `/send` as `messageId` and included in any delivery/bounce/complaint messages.

philbooth:
Depends on #72. Fixes #4.

Adds the ability to consume stored metadata from Redis and add it to the outgoing notification.

Again, no rush to review this, it can wait until after the All Hands.

### DIFF
--- a/src/message_data/mod.rs
+++ b/src/message_data/mod.rs
@@ -32,6 +32,18 @@ impl MessageData {
         }
     }
 
+    pub fn consume(&self, message_id: &str) -> Result<String, MessageDataError> {
+        let key = self.generate_key(message_id)?;
+        let key_str = key.as_str();
+        self.client
+            .get(key_str)
+            .map(|metadata| {
+                self.client.del::<&str, u8>(key_str).ok();
+                metadata
+            })
+            .map_err(From::from)
+    }
+
     pub fn set(&self, message_id: &str, metadata: &str) -> Result<(), MessageDataError> {
         let key = self.generate_key(message_id)?;
         self.client.set(key.as_str(), metadata).map_err(From::from)
--- a/src/message_data/test.rs
+++ b/src/message_data/test.rs
@@ -11,34 +11,70 @@ use sha2::Sha256;
 use super::*;
 use settings::Settings;
 
+#[derive(Debug)]
+struct TestFixture {
+    pub unhashed_key: String,
+    pub internal_key: String,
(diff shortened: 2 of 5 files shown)


src/message_data/mod.rs
```diff
@@ -32,6 +32,18 @@ impl MessageData {
         }
     }
 
+    pub fn consume(&self, message_id: &str) -> Result<String, MessageDataError> {
+        let key = self.generate_key(message_id)?;
+        let key_str = key.as_str();
+        self.client
+            .get(key_str)
+            .map(|metadata| {
+                self.client.del::<&str, u8>(key_str).ok();
```
vladikoff: My IDE freaks out over here: 
r

But seems like it's false to raise an error here?

src/message_data/mod.rs
```diff
@@ -32,6 +32,18 @@ impl MessageData {
         }
     }
 
+    pub fn consume(&self, message_id: &str) -> Result<String, MessageDataError> {
+        let key = self.generate_key(message_id)?;
```
vladikoff: I wonder if there is a way to to do in a single line for `key_str`....

src/message_data/test.rs
```diff
@@ -11,34 +11,70 @@ use sha2::Sha256;
 use super::*;
 use settings::Settings;
 
+#[derive(Debug)]
+struct TestFixture {
+    pub unhashed_key: String,
+    pub internal_key: String,
+    pub message_data: MessageData,
+    pub redis_client: RedisClient,
```
vladikoff: 👍

src/message_data/mod.rs
philbooth: Hmmm. Is your IDE set up to use nightly Rust? And have you done a `rustup update` recently?

philbooth: (that's probably not it, I'm just thinking out loud tbh)

vladikoff: It seems to be using latest Nightly in the IDE, but I will double check...

vladikoff: It's fine in VSCode , *shrugs*

Issue #48, opened 2018-06-05
Use failure crate for error handling


philbooth:
Recommendation from the push team. Example usage in [mozilla-services/megaphone](https://github.com/mozilla-services/megaphone) and [mozilla-services/pushbox](https://github.com/mozilla-services/pushbox).

philbooth:
(alternative recommendation was `error_chain`, but it sounds like `failure` is what the cool kids are using)

vladikoff:
@philbooth assigned to @brizental, she spent some time looking into error handling while working on the Logging PR

brizental:
I'll start looking at this now. I had a look at the megaphone error handling code and I think we can actually reuse a lot of their code to try this out.

PR #74, opened 2018-06-13
feat(logging): add mozlog to settings and scrub sensitive data

### LINKED ISSUE #10: Implement logging

### LINKED ISSUE #18: Scrub sensitive data when logging config
#17 adds sensitive data to the config. If you set those values in `config/local.json` or your environment, they'll be logged to the console. We should implement something to scrub them before this repo goes much further.

brizental:
Fixes #18 
Connects to #10 

I think this turned out nicer than my first trial to log the settings data :) Let me know what you guys think.

This is still a partial fix to #10 because I still need to look at logging for the queues.

This is how the settings log looks like:

```
{"Logger":"fxa-email-service-0.1.0","Type":"fxa-email-service:log","Pid":14300,"Severity":6,"Timestamp":1528911237258753000,"Fields":{"msg":"Finished configuration successfully.","settings":"{\"authdb\":{\"baseuri\":\"http://127.0.0.1:8000/\"},\"aws\":{\"keys\":\"[hidden]\",\"region\":\"us-east-1\",\"sqsurls\":null},\"bouncelimits\":{\"enabled\":true,\"complaint\":[{\"period\":86400000,\"limit\":0},{\"period\":31536000000,\"limit\":1}],\"hard\":[{\"period\":86400000,\"limit\":0},{\"period\":31536000000,\"limit\":1}],\"soft\":[{\"period\":300000,\"limit\":0}]},\"mozlog\":true,\"provider\":\"ses\",\"sender\":{\"address\":\"verification@latest.dev.lcip.org\",\"name\":\"Firefox Accounts\"},\"sendgrid\":\"[hidden]\"}"}}
```

---

[EDIT]

Since both queue and service use the settings function and the settings function now uses the logging module, the queue is now worrying about the service dependencies. That's not really nice...

I'll take a look at this and see if I can isolate each of them, instead of just importing everything from service into queue. That is also generating many warnings, which is annoying.

Should I open an issue for that or just try and fix it in this PR already? Do you guys have any idea on how to fix this easily?

Thanks!

---

r? @philbooth @vladikoff


philbooth:
> Since both queue and service use the settings function and the settings function now uses the logging module, the queue is now worrying about the service dependencies. That's not really nice...
>
> I'll take a look at this and see if I can isolate each of them, instead of just importing everything from service into queue. That is also generating many warnings, which is annoying.

Yeah, we definitely don't want *everything* from service in queue, but it should be fine if we just add this block to `queues_main.rs` I think?

```
#[macro_use(
    slog_o, slog_info, slog_kv, slog_log, slog_record, slog_b, slog_record_static, slog_error
)]
extern crate slog;
extern crate slog_async;
extern crate slog_mozlog_json;
extern crate slog_term;
```

> Should I open an issue for that or just try and fix it in this PR already?

Let's keep it in this PR. I can dig into it some more tomorrow morning if you haven't got it working before then.

src/app_errors/mod.rs
```diff
@@ -2,16 +2,13 @@
...
-use std::{collections::HashMap, result};
+use std::collections::HashMap;
 
-use failure::Error;
 use rocket_contrib::Json;
 
 #[cfg(test)]
 mod test;
 
-pub type AppResult<T> = result::Result<T, Error>;
```
philbooth: Is the plan to use `AppResult` more widely than just the logging module eventually? Might be worth keeping here if that's the case. (I confess I don't really get it what it's for at the moment)

brizental:
> Yeah, we definitely don't want everything from service in queue, but it should be fine if we just add this block to queues_main.rs I think?

I added that (and some more imports) but we are still getting some warnings of unused code when you run `cargo build`:

```
cargo b
   Compiling fxa-email-service v0.1.0 (file:///home/travis/build/mozilla/fxa-email-service)
warning: method is never used: `from_request`
  --> src/logging/mod.rs:34:5
   |
34 |     pub fn from_request(request: &Request) -> RequestMozlogFields {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(dead_code)] on by default
warning: method is never used: `with_request`
  --> src/logging/mod.rs:92:5
   |
92 |     pub fn with_request(request: &Request) -> AppResult<MozlogLogger> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    Finished dev [unoptimized + debuginfo] target(s) in 16.01s
```

src/app_errors/mod.rs
brizental: Yeah, so, that is needed for the logging and I think the plan is not to use it anywhere else. 

Since this code came from megaphone there are some decisions that they made that don't make sense to us, I guess. Probably they use this type in their error functions, but we just use it in the logging. That's why I moved it.

(I also don't get it 100%, I could investigate and try to understand it).

src/logging/mod.rs
```diff
@@ -53,7 +55,14 @@ impl KV for RequestMozlogFields {
         }
         serializer.emit_str("path", &self.path)?;
         serializer.emit_str("method", self.method)?;
+        Ok(())
+    }
+}
 
+impl Value for Settings {
```
philbooth: Would a more logical place for this be inside `src/settings/mod.rs`?

brizental: I wasn't sure about that, I left it in the logging because it's something for logging and this saves me some import lines. I don't have a strong opinion about this, though. If you think it's best I can change it :)

src/serialize.rs
```diff
@@ -0,0 +1,12 @@
+use serde::ser;
+
+pub fn hidden_or_not_set<T, S>(ref item: &Option<T>, serializer: S) -> Result<S::Ok, S::Error>
```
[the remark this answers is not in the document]
philbooth: ...hmm, no I don't, the thing I was thinking of below won't work.

We could just keep it simple and use `#[serde(skip_serializing)]` on those instead.

src/settings/mod.rs
```diff
@@ -122,8 +126,9 @@ impl Settings {
 
         match config.try_into::<Settings>() {
             Ok(settings) => {
-                // TODO: replace this with proper logging when we have it
-                println!("config: {:?}", settings);
+                let logger =
+                    MozlogLogger::new(&settings).expect("Unable to create MozlogLogger instance.");
+                slog_info!(logger, "{}", "Finished configuration successfully."; "settings" => &settings);
```
philbooth: Do we need to include a message where you have `Finished configuration successfully`? Seems like it's just log noise otherwise and this line is already going to be pretty long, so we might thank ourselves in the future if we ditch it or make it as short as possible.

philbooth:
@brizental you probably don't want to say "Partially fixes..." in the description because that is still going to close the named issue. I believe "Connects to..." is the official waffle-speak to get this positioned under that issue correctly, although it's also an annoyingly awkward piece of phrasing that I can't bring myself to use when I'm supposed to.

src/settings/mod.rs
brizental: Yes, I agree :) I'll remove that.

philbooth:
> I added that (and some more imports) but we are still getting some warnings of unused code when you run `cargo build`...

Ah yeah, of course. I think let's just ignore those for now, going forward this may be yet another black mark against my decision not to package stuff into a lib before splitting out the two processes. I'll open an issue to revisit that decision.

brizental:
> "Partially fixes..." in the description because that is still going to close the named issue. I believe "Connects to..." 

@philbooth Cool, I didn't know that. Will use the right terminology from now on.

src/serialize.rs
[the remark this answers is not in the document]
brizental: I guess to make this more generic we could just not specify that it's an `Option` and match `Some` and anything else to show `[hidden]`, while `None` will show `[not set]`. 

Also, I thought about using `#[serde(skip_serializing)]`, but then the user wouldn't know if the config was set or not and that might be useful information. With `skip_serializing` that config won't even show on the logs.

src/app_errors/mod.rs
pjenvey: AppResult isn't needed here currently since MozlogLogger::new here never returns an Error

megaphone's method can error out in a couple cases, one of them being determining the current hostname. You may also want to lookup the hostname similarly (MozLog standard recommends including one: https://wiki.mozilla.org/Firefox/Services/Logging )

src/serialize.rs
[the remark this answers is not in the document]
philbooth: > With `skip_serializing` that config won't even show on the logs.

Yep, but that's fine for stuff that isn't an `Option` because there is no "not set" state. The deserializer will fail before we can log the config if those aren't set.

vladikoff:
you gotta rebase! 👍

src/serialize.rs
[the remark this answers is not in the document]
brizental: This function was written with `Option` in mind. I guess if we wanted to scrub sensitive data from non-`Option` types we would just have to use a different approach. Like you said, if a value that is not an `Option` is not set, the whole thing would fail anyways. 

For this particular PR, everything that I want to hide is inside an `Option`, so I think we should think about hiding other types when we get there. For the `hmac_key` we can just use `#[serde(skip_serializing)]` or we could write some serializing function for those cases, something that will always return `[hidden]`, like:

```
pub fn hidden<T, S>(ref item: &T, serializer: S) -> Result<S::Ok, S::Error>
where
    S: ser::Serializer,
{
    serializer.serialize_str("[hidden]")
}
```

src/app_errors/mod.rs
brizental: I'll remove this for now and just use the normal `Result` type then. I'll take a closer look at error handling in #48 :)

src/serialize.rs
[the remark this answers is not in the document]
brizental: Alright! I understood you wrong, yeah, haha. Glad we're on the same page though :)

brizental:
Rebased @vladikoff :)

Also, changed the stuff that you asked on the review, @philbooth .

src/serialize.rs
```diff
@@ -0,0 +1,12 @@
+use serde::ser;
```
philbooth: This file needs the MPL preamble.

src/settings/mod.rs
```diff
@@ -5,64 +5,67 @@
 use std::env;
 
 use config::{Config, ConfigError, Environment, File};
+use logging::MozlogLogger;
```
philbooth: This `use` belongs in the block of internal stuff underneath.

src/settings/mod.rs
```diff
@@ -5,64 +5,67 @@
...
+#[derive(Debug, Default, Deserialize, Serialize)]
 pub struct AuthDb {
     #[serde(deserialize_with = "deserialize::base_uri")]
     pub baseuri: String,
 }
 
-#[derive(Debug, Default, Deserialize)]
+#[derive(Debug, Default, Deserialize, Serialize)]
 pub struct Aws {
+    #[serde(serialize_with = "serialize::hidden_or_not_set")]
```
philbooth: Can we add some test coverage to assert that all of these properties that are `hidden_or_not_set` get serialized correctly in both states?

philbooth:
@brizental, left a few more nits, with those done I reckon this will be good to merge! :+1:

vladikoff:
@philbooth final r? 

(Side note: could you review the existing issues and comment in them where we need @brizental 's attention, I feel like we are stuck not-coding new stuff and lost pace)

brizental:
@philbooth just added these last changes and the tests. 

I realize I did the tests a little differently than what you have been doing with your deserialize functions so let me know what you think of them :)

src/serialize/test.rs
```diff
@@ -0,0 +1,48 @@
...
+            Token::Struct {
+                name: "TestStruct",
+                len: 1,
+            },
+            Token::Str("a"),
+            Token::Str("[hidden]"),
+            Token::StructEnd,
+        ],
+    );
+}
```
philbooth: These tests are fine as far as they go but they don't fail if somebody accidentally deletes the `hidden_or_not_set` from one of the sensitive fields, which is the thing I really want to protect against (as it seems far more likely to happen than someone accidentally breaking `hidden_or_not_set` itself.

So the kind of tests I'd really like to see would go in `src/settings/test.rs` and would assert that the serialized string for `Settings` does not contain whatever preset values you assign for the sensitive fields. I've not tried actually writing that out, but it seems like it should be straightforward enough in practice?

philbooth:
> could you review the existing issues and comment in them where we need @brizental 's attention

I'm not sure what kind of comments you're after here tbh. Everything that is not already assigned is fair game, @brizental feel free to pick up whatever you want to work on and let me know if anything written in an issue doesn't make sense or needs more explanation.

> I feel like we are stuck not-coding new stuff and lost pace

Sorry about that, I have some other stuff going on right now that's more important.

vladikoff:
> Everything that is not already assigned is fair game, @brizental feel free to pick up whatever you want to work

👍 this sounds good, I was mostly looking for some pointers to what is more urgent to finish, so we can maximize our assistance with the project. I apologize if it sounded too intense. We should be good for now with the rest of the existing bugs. Thanks Phil!

src/serialize/test.rs
brizental: Makes a lot of sense! I'll check that out.

brizental:
@philbooth I added the tests, but just for the `[hidden]` case. 

I think that is fine since we are testing the serializing function already, but I did that because the `[not set]` will have unpredictable behaviour, since our `CleanEnvironment` only cleans the env variables, but doesn't change the `local.json` and `dev.json` stuff. 

Basically, if I just clean the environment variables but still have aws and sendgrid set in the `local.json` or `dev.json` the tests will fail because they will create a settings with the keys.

I think that might be an issue for later. Maybe we should have a `test.json` for testing variables or not take the `local.json` or `dev.json` into account when testing.

Cargo.toml
```diff
@@ -24,7 +22,12 @@ sendgrid = ">=0.7.0"
 serde = ">=1.0"
 serde_derive = ">=1.0"
 serde_json = ">=1.0"
+serde_test = "1.0.66"
```
philbooth: Tiny nit, can we use `>=` here and keep `chrono` in the main depenency list?

src/settings/test.rs
```diff
@@ -185,6 +187,36 @@ fn env_vars_take_precedence() {
...
+    match Settings::new() {
+        Ok(settings) => {
+            let json = serde_json::to_string(&settings).unwrap();
+            let s: Value = serde_json::from_str(&json).unwrap();
+            assert_eq!(s["sendgrid"], "[hidden]");
+            assert_eq!(s["aws"]["keys"], "[hidden]");
+        }
+        Err(_error) => assert!(false, "Settings::new shouldn't have failed"),
+    }
+}
```
philbooth: Perfect, thanks!

PR #80, opened 2018-06-17
refactor(db): hide queues db access behind bounces interface

### LINKED ISSUE #68: Hide auth db access behind the bounces module in queues process
The `bounces` module should be our acceptable interface over the auth db and all db access should pass through it. We're doing that in the send service, but not in the queues.

philbooth:
Fixes #68.

We don't want to let knowledge about the auth db leak into any unrelated application logic, because it's both temporary and in possession of capabilities that this repo has no business knowing about.

@mozilla/fxa-devs r?

### DIFF
--- a/src/auth_db/mod.rs
+++ b/src/auth_db/mod.rs
@@ -4,7 +4,7 @@
 
 use std::{
     error::Error,
-    fmt::{self, Display, Formatter},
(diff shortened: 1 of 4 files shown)


vladikoff:
@brizental r?

PR #85, opened 2018-06-19
refactor(errors): use failure crate to handle errors

brizental:
Relates to #48 

This is still a work in progress, but you guys can already have a look at how this would work out.

### DIFF
--- a/src/app_errors/mod.rs
+++ b/src/app_errors/mod.rs
@@ -2,65 +2,38 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
-use std::collections::HashMap;
 
-use rocket_contrib::Json;
+use error::{HandlerErrorKind, HandlerResult};
 
-#[cfg(test)]
-mod test;
+// #[cfg(test)]
+// mod test;
 
 #[error(400)]
-pub fn bad_request() -> Json<ApplicationError> {
-    Json(ApplicationError::new(400, "Bad Request"))
+pub fn bad_request() -> HandlerResult<()> {
+    Err(HandlerErrorKind::BadRequest)?
 }
 
 #[error(404)]
-pub fn not_found() -> Json<ApplicationError> {
-    Json(ApplicationError::new(404, "Not Found"))
+pub fn not_found() -> HandlerResult<()> {
+    Err(HandlerErrorKind::NotFound)?
 }
 
 #[error(405)]
-pub fn method_not_allowed() -> Json<ApplicationError> {
-    Json(ApplicationError::new(405, "Method Not Allowed"))
+pub fn method_not_allowed() -> HandlerResult<()> {
+    Err(HandlerErrorKind::MethodNotAllowed)?
 }
 
 #[error(422)]
-pub fn unprocessable_entity() -> Json<ApplicationError> {
-    Json(ApplicationError::new(422, "Unprocessable Entity"))
+pub fn unprocessable_entity() -> HandlerResult<()> {
+    Err(HandlerErrorKind::UnprocessableEntity)?
 }
 
 #[error(429)]
-pub fn too_many_requests() -> Json<ApplicationError> {
-    Json(ApplicationError::new(429, "Too Many Requests"))
+pub fn too_many_requests() -> HandlerResult<()> {
+    Err(HandlerErrorKind::TooManyRequests)?
 }
 
 #[error(500)]
-pub fn internal_server_error() -> Json<ApplicationError> {
-    Json(ApplicationError::new(500, "Internal Server Error"))
-}
-
-#[derive(Debug, Deserialize, PartialEq, Serialize)]
-pub struct ApplicationError {
-    pub status: u16,
-    pub error: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub errno: Option<u8>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub message: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub data: Option<HashMap<String, String>>,
-}
-
-impl ApplicationError {
-    pub fn new(status: u16, error: &str) -> ApplicationError {
-        // TODO: Set errno, message and data when rocket#596 is resolved
-        //       (https://github.com/SergioBenitez/Rocket/issues/596)
-        ApplicationError {
-            status,
-            error: error.to_string(),
-            errno: None,
-            message: None,
-            data: None,
-        }
-    }
+pub fn internal_server_error() -> HandlerResult<()> {
+    Err(HandlerErrorKind::InternalServerError)?
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,125 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::fmt;
+use std::result;
+
+use failure::{Backtrace, Context, Fail};
+use rocket::http::{Status};
+use rocket::response::{Responder, Response};
+use rocket::{self, response, Request};
+use rocket_contrib::Json;
+
+use logging::MozlogLogger;
+
+pub type HandlerResult<T> = result::Result<T, HandlerError>;
+
+#[derive(Debug)]
+pub struct HandlerError {
+    inner: Context<HandlerErrorKind>,
+}
+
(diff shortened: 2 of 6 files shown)


src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        let json = Json(json!({
+            "status": status.code,
+            "error": format!("{}", self)
+        }));
+
+        let mut builder = Response::build_from(json.respond_to(request)?);
+        builder.status(status).ok()
+    }
+}
```
vladikoff: new line end of file

src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+    #[fail(display = "Unprocessable Entity")]
+    UnprocessableEntity,
+
+    /// 429 Too Many Requests
+    #[fail(display = "Too Many Requests")]
+    TooManyRequests,
+
+    /// 500 Internal Server Error
+    #[fail(display = "Internal Server Error")]
+    InternalServerError,
```
vladikoff: Is there a way to provide detailed logging in DEV mode for `InternalServerError`s?

src/send/test.rs
```diff
@@ -9,10 +9,11 @@ use rocket::{
 };
 use serde_json;
 
-use app_errors::{self, ApplicationError};
+use app_errors;
 use auth_db::DbClient;
 use bounces::Bounces;
 use message_data::MessageData;
+use error::{HandlerErrorKind};
```
vladikoff: do we need alphabetical order for this?

src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    #[fail(display = "Invalid Email Address")]
+    InvalidEmailAddress,
+
+    #[fail(display = "Unexpected fxa-email-service error")]
+    InternalError,
+
+    #[fail(display = "Unexpected rocket error: {:?}", _0)]
+    RocketError(rocket::Error), // rocket::Error isn't a std Error (so no #[cause])
```
vladikoff: `a std Error` -> `an std Error`

src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+pub type HandlerResult<T> = result::Result<T, HandlerError>;
+
+#[derive(Debug)]
+pub struct HandlerError {
+    inner: Context<HandlerErrorKind>,
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Fail)]
+pub enum HandlerErrorKind {
+    /// 400 Bad Request
```
vladikoff: do we need to handle `401`s ? (`401 unauthorized`)

src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+/// Generate HTTP error responses for HandlerErrors
+impl<'r> Responder<'r> for HandlerError {
+    fn respond_to(self, request: &Request) -> response::Result<'r> {
+        let status = self.kind().http_status();
+        let log = MozlogLogger::with_request(request).map_err(|_| Status::InternalServerError)?;
+        let kind = self.kind();
+        slog_error!(log, "{}", &self; "status_code" => status.code,  "status_msg" => status.reason);
```
vladikoff: Based on 
```
-            status,
-            error: error.to_string(),
-            errno: None,
-            message: None,
-            data: None,
```

Can we provide `errno` and maybe other fields? I think that's the field format that we should into mozlog

philbooth: ...and if we do do that (which I agree with), does it make sense to do it via `{:?}` / implementing `fmt:::Debug`?

I also wonder if, for certain kinds of error, we should include `inner.backtrace()` in the debug output, as that could be one way to address https://github.com/mozilla/fxa-email-service/pull/85/files#r196545966 (although maybe not the optimal way).

src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+
+pub type HandlerResult<T> = result::Result<T, HandlerError>;
+
+#[derive(Debug)]
+pub struct HandlerError {
+    inner: Context<HandlerErrorKind>,
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Fail)]
+pub enum HandlerErrorKind {
```
philbooth: It's no biggie but fwiw I'm not a huge fan of the `...Kind` suffix in this name. Is it a Rust convention or just something we opted for ourselves?

src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use failure::{Backtrace, Context, Fail};
+use rocket::http::{Status};
+use rocket::response::{Responder, Response};
+use rocket::{self, response, Request};
+use rocket_contrib::Json;
+
+use logging::MozlogLogger;
+
+pub type HandlerResult<T> = result::Result<T, HandlerError>;
```
philbooth: Does `Result` need to be namespace-qualified here? Isn't it included in the standard prelude?

src/send/mod.rs
```diff
@@ -75,8 +76,11 @@ fn validate(email: &Email) -> bool {
     true
 }
 
-fn fail() -> data::Outcome<Email, ()> {
-    Outcome::Failure((Status::BadRequest, ()))
+fn fail() -> data::Outcome<Email, HandlerError> {
+    Outcome::Failure((
+        Status::BadRequest, 
+        HandlerErrorKind::InvalidEmailAddress.into()
```
philbooth: This probably shouldn't be hard-coded at this point because `fail` is also called if the `provider` param is invalid. And maybe it will be called for other params in the future, too. Let's either rework the outer logic so the `HandlerErrorKind` can be passed in as an argument, or keep that the same and change this to a more generic validation error that doesn't mention email addresses.

src/app_errors/mod.rs
```diff
@@ -2,65 +2,38 @@
...
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub message: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub data: Option<HashMap<String, String>>,
-}
-
-impl ApplicationError {
-    pub fn new(status: u16, error: &str) -> ApplicationError {
-        // TODO: Set errno, message and data when rocket#596 is resolved
-        //       (https://github.com/SergioBenitez/Rocket/issues/596)
```
philbooth: I'd like it if we could keep some form of this `TODO` comment in the code, as a visible reminder that we have outstanding work to do in order to make the errors from this service match our other errors (and what was specced out in the feature doc).

philbooth: Presumably that just means moving it down to the `HandlerError` definition. If we do that maybe we should also keep the placeholders for `errno`, `message` and `data` there too, because we know we have to implement those.

src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+
+#[derive(Debug)]
+pub struct HandlerError {
```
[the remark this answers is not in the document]
philbooth: (and if we do go back to the name `ApplicationError`, this stuff should probably stay in `app_errors/mod.rs`)

src/error.rs
philbooth: Not yet, there's no authentication in the first cut of this service, so we never return a 401.

src/error.rs
vladikoff: My mistake, ` a standard Error` is right 👍

src/error.rs
brizental: Yes, that's what I'm going for :)

src/send/test.rs
brizental: This is still WIP, so I haven't use `cargo fmt -- --check` yet, that will tell me about this kind of formatting errors.

src/error.rs
brizental: Hm, I followed this tutorial from the `failure` book -> https://boats.gitlab.io/failure/error-errorkind.html

It's the naming convention that they use there, but I guess we could name it something we think is more semantic. What did you have in mind?

src/error.rs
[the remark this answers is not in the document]
brizental: I agree with going back to `ApplicationError` now that I think about it, It is is more semantic.

src/error.rs
philbooth: No worries, if it's an idiomatic Rust thing I'm happy!

src/error.rs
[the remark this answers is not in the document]
pjenvey: I called them HandlerError/Result because ultimately rocket handlers return them as HTTP responses (they impl. rocket's Responder). Though maybe one day [rocket's request guards will also render them as Responders for us (Rocket#596)](https://github.com/SergioBenitez/Rocket/issues/596) -- TBD what that will ultimately look like.

If you really want to change it I'd recommend a shorter "AppError". We'll probably just stick w/ "Handler" in megaphone/pushbox until Rocket#596 shakes out

brizental:
Closing this in favour of #94

PR #86, opened 2018-06-19
fix(validation): make email regex less restrictive

### LINKED ISSUE #81: Investigate the dot issue in emails, curl fails for `some.some@gmail.com`

brizental:
Fixes #81 

Inspired by: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/validate.js#L30
Tested with: https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/lib/validate.js#L18 and also https://blogs.msdn.microsoft.com/testing123/2009/02/06/email-address-test-cases/

r? @vladikoff @philbooth 

Should we add test cases for this?


vladikoff:
> Should we add test cases for this?

Can we add all test cases from fxa-content-server ?

vladikoff:
@rfk also see https://github.com/mozilla/fxa-email-service/issues/9

src/validate/mod.rs
```diff
@@ -15,7 +15,7 @@ lazy_static! {
         "^https?://[A-Za-z0-9-]+(?:\\.[A-Za-z0-9-]+)*(?::[0-9]+)?/(?:[A-Za-z0-9-]+/)*$"
     ).unwrap();
     static ref EMAIL_ADDRESS_FORMAT: Regex = Regex::new(
-        "^[a-z0-9\\-!#\\$%\\&`\\*\\+/=\\?\\^{\\|}~]+@[a-z0-9-]+(?:\\.[a-z0-9-]+)+$"
+        "^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,253}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,253}[a-zA-Z0-9])?)+$"
```
philbooth: If we're going to permit uppercase in the API, we need to make sure that we normalise the case before getting and setting bounce records in the db. Otherwise a bounce to `FOO@example.com` will not be matched by a query against `foo@example.com`.

brizental: Alright, should I just fix it in this PR or should we open another issue about this? 

I feel this should be alright, Rust has a `to_lowercase` method (https://doc.rust-lang.org/std/string/struct.String.html#method.to_lowercase).

philbooth: If you're happy with it, I think we should just fix it in this PR.

vladikoff: Also need ba test for it

brizental: Hey guys, so I thought about just adding the `to_lowercase` function to to the bounces code, but it seemed cleaner to just make the email lowercase on deserialize (https://github.com/mozilla/fxa-email-service/pull/86/commits/37e8bbb8ae222930ba8c27bc976f2e4924107ac8#diff-170f5e597783d14eb65abe649bb9d4d2R45). 

Let me know if my thinking was right :)

src/deserialize.rs
```diff
@@ -41,7 +41,8 @@ pub fn email_address<'d, D>(deserializer: D) -> Result<String, D::Error>
 where
     D: Deserializer<'d>,
 {
-    deserialize(deserializer, validate::email_address, "email address")
+    let email = deserialize(deserializer, validate::email_address, "email address")?;
+    Ok(email.to_lowercase())
```
vladikoff: >  seemed cleaner to just make the email lowercase on deserialize

Is it possible to add a test for `UPPERCASE@emails.com` ?

brizental: Done.

brizental: Done.

philbooth: Just noting for future reference that although making it lower case at this point is fine for now, longer term we may want to think about keeping it internal to the `Bounces` module.

These email addresses get reported back to the caller via the outgoing notification queue and it might be a less surprising API if the email addresses that go out can be directly string-compared to the ones that were passed in. For the auth server this probably isn't an issue because we'll call it with `normalizedEmail` anyway. But if/when we're open to other callers it might be something to think about.

vladikoff: >  But if/when we're open to other callers it might be something to think about.

👍

Issue #75, opened 2018-06-13
Add headers option to /send POST params


brizental:
This came up when I was looking at #46 .

We add some custom headers to our emails (https://github.com/mozilla/fxa-auth-server/blob/master/test/mail_helper.js#L42-L48), and right now we don't have an option to do that with this service. The new option would look like this: 

```
curl \
  -d '{"to":"foo@example.com","subject":"bar","headers": {"x-whatever": "blablabla"},"body":{"text":"baz"}}' \
  -H 'Content-Type: application/json' \
  http://localhost:8001/send
```

I sent a PR to the sendgrid-rs package, adding an `add_headers` function to the personalizations (https://github.com/gsquire/sendgrid-rs/pull/26), so that should be easy enough.

The SES though, would need to be refactored to use `SendRawEmailRequest` (https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-raw.html) instead of `SendEmailRequest`, which is what we are using right now and doesn't accept custom headers.

philbooth:
@brizental I don't think we need to worry about this.

Afaik, those headers are only used as a hack for propagating state to the bounce/complaint/delivery notifications. In this repo we are replacing it with proper storage keyed by the message id, as per #72 and #73. Although maybe @vbudhram or @vladikoff can set me straight if there are other legitimate uses for them.

Good thinking though! :smile:

philbooth:
>  we send template names

Can't those go in Redis as part of the metadata blob?

> we needed for the STDOUT provider

Why though? That provider can have a delivery queue stub that works just like the other providers, reading data from Redis, can't it?

vladikoff:
@philbooth don't we use those like everywhere: https://github.com/mozilla/fxa-auth-server/blob/677bdbb6a80da2f4160cf47fc129eb5a43b9f8ee/test/mail_helper.js#L42 I find them very useful, we don't have to include the bounce headers hacks in this though

philbooth:
I'd prefer if there was just one mechanism for associating data with a message, because it makes both the code and the API simpler. For the mail helper, that would mean there is some refactoring to do for it to use the API we've specced out here. The alternative, of refactoring here so that the mail helper can remain unchanged feels like a poor tradeoff to me. A third option, which I'd also be comfortable with, is just to keep the mail helper completely separate to this repo and don't both with adding a logging provider implementation here.

[...]

Issue #21, opened 2018-05-08
Write tests for the SES provider


philbooth:
Currently the tests all use the `mock` provider. We should figure out how to mock the framework around Rusoto and then write tests that exercise the `ses` provider too.

brizental:
I was thinking about which tests we need here, I'm not really sure about it so, these are the ones I came up with, if you think we need more than these or if they are completely wrong, lmk.

- Check that RawMessage is created correctly for these cases:
   - with | without headers
   - cc.len() > 0 | cc.len() < 0

philbooth:
@brizental, that sounds about right to me!

PR #89, opened 2018-06-21
feat(providers): add custom headers option

brizental:
Relates to #75 

This still needs tests, but I thought we should discuss which tests together :)


Cargo.toml
```diff
@@ -23,7 +25,7 @@ rusoto_core = ">=0.32.0"
 rusoto_credential = ">=0.11.0"
 rusoto_ses = ">=0.32.0"
 rusoto_sqs = ">=0.32.0"
-sendgrid = ">=0.7.0"
+sendgrid = { git = "https://github.com/gsquire/sendgrid-rs.git" }
```
vladikoff: 0.7.1 is already available with your patch https://crates.io/crates/sendgrid/0.7.1 right?

src/providers/ses.rs
```diff
@@ -44,44 +46,70 @@ impl Provider for SesProvider {
...
+
+        let encoded_message = encode(&format!("{}", message));
+        let mut request = SendRawEmailRequest::default();
+        request.raw_message = RawMessage {
+            data: format!("{}", encoded_message).as_bytes().to_vec(),
+        };
 
         self.client
-            .send_email(&request)
+            .send_raw_email(&request)
```
vladikoff: with the amount of changes to switch to RawMessage to support headers I think we should land this PR and then spin up a different issue / PR to finalize the ALLOW_LIST passthrough for the known headers (from https://github.com/mozilla/fxa-email-service/issues/75#issuecomment-398420128 )

vladikoff:
> This still needs tests, but I thought we should discuss which tests together :)

I wonder if we have the technology to mock out the <kbd>self.client.send</kbd> and <kbd>self.client.send_raw_message</kbd> functions to later assert that the properties of the `&request`. So we can later check the contents of the `headers` 🤔 

At a quick glance I didn't see any tests for `ses` and `sendgrid` providers in the `/providers` directory. 
image

Cargo.toml
brizental: ahh yes that's right :)

src/providers/ses.rs
brizental: yes that's what i was thinking.

brizental:
> At a quick glance I didn't see any tests for ses and sendgrid providers in the /providers directory.

@vladikoff there is an issue open for that ->  #21

vladikoff:
> @vladikoff there is an issue open for that -> #21

Well here we go ❗️

src/providers/ses.rs
```diff
@@ -44,44 +46,70 @@ impl Provider for SesProvider {
...
         cc: &[&str],
+        headers: Option<&Headers>,
         subject: &str,
         body_text: &str,
         body_html: Option<&str>,
     ) -> Result<String, ProviderError> {
-        let mut destination = Destination::default();
-        destination.to_addresses = Some(vec![to.to_string()]);
+        let mut all_headers = header::Headers::new();
+        all_headers.set(header::From(vec![self.sender.parse().unwrap()]));
```
philbooth: Panicking inside a route handler is not acceptable I think. If one of these `parse` calls fails, we should fail gracefully and send an appropriate error response back to the caller.

brizental:
@vladikoff @philbooth hey guys, I just sent in a commit for the reviews you asked for. r?

src/providers/ses.rs
```diff
@@ -44,44 +46,70 @@ impl Provider for SesProvider {
...
-            message.body.html = Some(Content {
-                charset: None,
-                data: html.to_string(),
-            });
+        let mut message: Message<MultiPart<String>> = Message::new().with_headers(all_headers);
+
+        let mut body: MultiPart<String> = MultiPart::alternative().with_singlepart(
+            SinglePart::quoted_printable()
+                .with_header(header::ContentType(
+                    "text/plain; charset=utf8".parse().unwrap(),
```
philbooth: Just noting for future reference, when we open this API up to external callers we should make it explicit in the docs that UTF-8 is assumed (or provide a param that lets them override it).

PR #94, opened 2018-06-23
refactor(errors): use failure crate to handle errors

brizental:
Connects to #48 

I had to change basically everything since my last PR (#85) about this, so I thought it would be best to just create a new PR.

To  get the failure errors working I needed to get at least `providers`, `bounces` and `db` to also work with the failure errors, not just the Rocket and HTTP errors. That's what I'm doing in this PR. Since this is kind of a big change, I thought it would be best to send this in before adapting the tests to the new error type, so, right now, many tests are commented.

Let me know what you guys think. If you think this is a good change I still would need to adapt all tests and also there are some error types specially for the queues bin that need to be changed as well.

I personally think this is a good change. It is abstracting all the errors to one single place in the code. While working on this refactoring, I saw a bunch of repeated code for error handling. This ends that, which means it will be easier to maintain and create new error types with this way of doing things. Also, we are now getting much more information about each error, not just the HTTP status and message and it's very easy to customize this even further.

Anyways, this still has a bunch of work to be ready to merge :)

r? @vladikoff @philbooth @pjenvey

### DIFF
--- a/src/app_errors/mod.rs
+++ b/src/app_errors/mod.rs
@@ -2,65 +2,284 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
-use std::collections::HashMap;
+use std::{fmt, result};
 
+use failure::{Backtrace, Context, Fail};
+use rocket::{
+    self,
+    http::Status,
+    response::{self, Responder, Response},
+    Request,
+};
 use rocket_contrib::Json;
+use serde_json::{map::Map, ser::to_string, Value};
+
+use auth_db::BounceRecord;
+use logging::MozlogLogger;
 
 #[cfg(test)]
 mod test;
 
+pub type AppResult<T> = result::Result<T, AppError>;
+
+#[derive(Debug)]
+pub struct AppError {
+    inner: Context<AppErrorKind>,
+}
+
+impl AppError {
+    pub fn json(&self) -> Value {
+        let kind = self.kind();
+        let status = kind.http_status();
+
+        let mut fields = Map::new();
+        fields.insert(
+            String::from("code"),
+            Value::String(format!("{}", status.code)),
+        );
+        fields.insert(
+            String::from("error"),
+            Value::String(format!("{}", status.reason)),
+        );
+        let errno = kind.errno();
+        if let Some(ref errno) = errno {
+            fields.insert(String::from("errno"), Value::String(format!("{}", errno)));
+            fields.insert(String::from("message"), Value::String(format!("{}", self)));
+        };
+        let additional_fields = kind.additional_fields();
+        for (field, value) in additional_fields.iter() {
+            fields.insert(field.clone(), value.clone());
+        }
+
+        json!(fields)
+    }
+
+    pub fn kind(&self) -> &AppErrorKind {
+        self.inner.get_context()
+    }
+}
+
+impl Fail for AppError {
+    fn cause(&self) -> Option<&Fail> {
+        self.inner.cause()
+    }
+
+    fn backtrace(&self) -> Option<&Backtrace> {
+        self.inner.backtrace()
+    }
+}
+
+impl fmt::Display for AppError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.inner.fmt(f)
+    }
+}
+
+/// The specific kind of error that can occur.
+#[derive(Clone, Debug, Eq, Fail, PartialEq)]
+pub enum AppErrorKind {
+    /// 400 Bad Request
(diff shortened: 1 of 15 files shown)


src/error.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// 404 Not Found
+    NotFound,
+    /// 405 Method Not Allowed
+    MethodNotAllowed,
+    /// 422 Unprocessable Entity
+    UnprocessableEntity,
+    /// 429 Too Many Requests,
+    TooManyRequests,
+    /// 500 Internal Server Error
+    InternalServerError,
```
philbooth: Seeing both these and our internal error types modelled inside `AppErrorKind` gives me a confusing mental model. Do we need to model the HTTP error types here, or do they all just fall out naturally if we do the appropriate work when serializing/mapping our internal types?

(not sure if that's clear, I mean if the implementations of `fmt` and `http_status` are correct for all our internal errors and we only return those internally, would they ever get invoked with these values?)

src/error.rs
```diff
@@ -0,0 +1,186 @@
...
+impl<'r> Responder<'r> for AppError {
+    fn respond_to(self, request: &Request) -> response::Result<'r> {
+        let status = self.kind().http_status();
+        let log = MozlogLogger::with_request(request).map_err(|_| Status::InternalServerError)?;
+        slog_error!(log, "{}", &self; "status_code" => status.code,  "status_msg" => status.reason);
+
+        let json = Json(json!({
+            "kind": self.kind().as_ref(),
+            "status": status.code,
+            "message": format!("{}", self)
```
philbooth: We have a standard, pre-existing format for these error payloads, described in the feature doc and also used by the other FxA services (see e.g. `lib/error.js` in `fxa-auth-server` for our definition of `AppError` over there). Let's try to get as close to that as possible.

So `status` should be `code`. Maybe we can map `kind` to `errno`? And we should have separate `error` and `message` fields, where `error` is the standard HTTP error string for each status, and `message` is our more detailed string for the specific error.

src/error.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl From<Context<AppErrorKind>> for AppError {
+    fn from(inner: Context<AppErrorKind>) -> AppError {
+        AppError { inner }
+    }
+}
+
+/// Generate HTTP error responses for AppErrors
+impl<'r> Responder<'r> for AppError {
+    fn respond_to(self, request: &Request) -> response::Result<'r> {
```
philbooth: I'm just thinking out loud here, so we don't need to worry for this PR, but seeing this suggests that maybe in time we might want to have different implementations of this for error payloads that contain extra fields (e.g. validation errors where ultimately we want to list the invalid params etc). Or maybe we can do all that in this one implementation by having some kind of generic extra-fields mechanism. Dunno.

src/error.rs
```diff
@@ -0,0 +1,186 @@
...
+};
+use rocket_contrib::Json;
+
+use auth_db::BounceRecord;
+use logging::MozlogLogger;
+
+pub type AppResult<T> = result::Result<T, AppError>;
+
+#[derive(Debug)]
+pub struct AppError {
```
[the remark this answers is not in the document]
brizental: Makes sense to just put them all together, yes. In the beginning I thought it didn't and then I just forgot to get them all together again.

src/error.rs
brizental: Hmm, I understand what you're saying. I didn't really think of that, but it makes sense.

I'm not sure about what is the answer for this right now, but I will investigate and let you know.

brizental: I tried removing these handlers, but what happens is that it falls back to the default rocket handler for them and we get a HTML response and no logging. I'll investigate this a little further, but I think we have to have these, even though they seem redundant.

brizental:
Hey guys, so I added a few changes to this. I'll try and list them all here:

- Divided `BounceError` into `BounceComplaintError`, `BounceSoftError`, `BounceHardError`.
- Added `errno` fields to all custom errors types. HTTP errors don't get `errno`s.
- Changed the fields in the log message to be just like the fields in `fxa-auth-server`. (https://github.com/mozilla/fxa-auth-server/blob/master/lib/error.js#L78-L81)
- Added the option to have extra fields for some the custom errors. eg. bounces will have a `bounce` and `address` field in their log message or providers will have a `name` field in their error message (let me know if you think the way I did it was too crazy, I was not too sure about it).

What still needs to be done is:

- Get all tests to pass.
- Maybe add more additional fields to some of the error if needed.
- Document the errors and `errno`s.

philbooth:
@brizental, let's not worry about these two in this PR, they're both peripheral to the substantive change here:

> * Maybe add more additional fields to some of the error if needed.
> * Document the errors and errnos.

PR #102, opened 2018-06-26
feat(logging): add null logger for testing environment

### LINKED ISSUE #82: Create a null logger for use in the tests
Currently the test output is interleaved with a load of settings json from all the tests that instantiate `Settings`. We should try to come up with some kind of null logger that we can specify from the tests, so that the test output is easier to read.

brizental:
Fixes #82 

When we are in a testing environment (which means when `NODE_ENV=test`) the logger won't show anything. 

So now, instead of having `mozlog:true | false` in the config, we have `logging: null | pretty | mozlog`.

r? @vladikoff

### DIFF
--- a/config/test.json
+++ b/config/test.json
@@ -1,0 +1,3 @@
+{
+    "logging": "null"
+}
--- a/scripts/test_with_authdb.sh
+++ b/scripts/test_with_authdb.sh
@@ -17,5 +17,9 @@ fi
 
 sleep 2
 
+if [ -z "$NODE_ENV" ]; then
+  export NODE_ENV=test
+fi
+
 export RUST_BACKTRACE=1
 cargo test -- --test-threads=1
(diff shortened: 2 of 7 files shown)


config/test.json
```diff
@@ -0,0 +1,3 @@
+{
+    "logging": "null"
```
vladikoff: @philbooth thoughts on the name `null` ?

scripts/test_with_authdb.sh
```diff
@@ -17,5 +17,7 @@ fi
 
 sleep 2
 
+if [ -z "$NODE_ENV" ]; then export NODE_ENV=test; fi;
```
philbooth: Tiny nit, but lets be liberal with newline characters in these scripts, we have plenty spare.

config/test.json
philbooth: I like it.

scripts/test_with_authdb.sh
brizental: Done.

PR #103, opened 2018-06-26
Release v1.115.0

vladikoff:
An initial release to sync up with the rest of the projects. The steps are a bit manual

I looked at more automation with https://github.com/sunng87/cargo-release but not done yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,50 @@
+<a name="1.115.0"></a>
+## 1.115.0 (2018-06-26)
+
+
+#### Bug Fixes
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,50 @@
+<a name="1.115.0"></a>
```
philbooth: As we're starting from scratch / doing it manually, is it okay if we fix something that's always bugged me in the other repos? These anchors are completely redundant, GitHub does it all automagically from the heading anyway.

philbooth: Oh nvm, it's generated I notice.

PR #104, opened 2018-06-27
chore(settings): prepare prod and stage configs for deployment

### LINKED ISSUE #24: Set rocket config for stage and prod

### LINKED ISSUE #100: Do not run in production without a changed message_id_hmac_key
We should panic or something if we run the service in production and it does not the `message_id_hmac_key` value changed

brizental:
Fixes #100 
Fixes #24 

As mentioned in #24 we still need to see about `ROCKET_WORKERS`.

Also, in production, rocket sends a warning about not having set a secret_key, but I decided to leave it without anything for now. Anyways that would be something that probably we wouldn't "hard code" here.

r? @vladikoff @philbooth

### DIFF
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -130,11 +130,18 @@ impl Settings {
         }
 
         config.merge(File::with_name("config/local").required(false))?;
-
         config.merge(Environment::with_prefix("fxa_email"))?;
 
         match config.try_into::<Settings>() {
             Ok(settings) => {
+                if let Ok(rocket_env) = env::var("ROCKET_ENV") {
+                    if rocket_env == "production"
+                        && &settings.message_id_hmac_key == "YOU MUST CHANGE ME"
+                    {
+                        panic!("Please set a valid HMAC key.")
+                    }
+                }
+
                 let logger =
                     MozlogLogger::new(&settings).expect("Unable to create MozlogLogger instance.");
                 slog_info!(logger, "Settings::new"; "settings" => &settings);
--- a/Rocket.toml
+++ b/Rocket.toml
@@ -1,2 +1,11 @@
 [development]
+address = "127.0.0.1"
+port = 8001
+
+[staging]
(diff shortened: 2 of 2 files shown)


src/settings/mod.rs
```diff
@@ -130,11 +130,18 @@ impl Settings {
...
 
         config.merge(File::with_name("config/local").required(false))?;
-
         config.merge(Environment::with_prefix("fxa_email"))?;
 
         match config.try_into::<Settings>() {
             Ok(settings) => {
+                if let Ok(rocket_env) = env::var("ROCKET_ENV") {
+                    if rocket_env == "production"
+                        && &settings.message_id_hmac_key == "YOU MUST CHANGE ME"
```
brizental: I'm thinking here it would be best if we actually checked that the HMAC was valid, instead of just checking for `YOU MUST CHANGE ME`. But I remember us talking in a meeting about doing it this way, so let me know what you guys think :)

vladikoff: > 'm thinking here it would be best if we actually checked that the HMAC was valid

Sure let's do that if possible, and we also need a sample one for `dev.json` right?

brizental: When something is not set in `dev.json` it defaults to `default.json` so in there we would just get `YOU MUST CHANGE ME`. Do you think we should have something else for `dev.json`?

vladikoff: from mtg: So these can be any string so let's keep it this way

PR #105, opened 2018-06-27
feat(docker): set ROCKET_ENV to prod

vladikoff:


Dockerfile
```diff
@@ -35,4 +35,6 @@ COPY --from=builder /app/bin /app/bin
 WORKDIR /app/bin
 USER app
 
+ENV ROCKET_ENV production
```
vladikoff: this guards production from failing to set `ROCKET_ENV`.

Override works with something like `docker run -e "ROCKET_ENV=development"  mozilla/fxa-email-service:feature.prodimage`

PR #110, opened 2018-06-28
fix(settings): get message data hmac key setting to work with env variables

### LINKED ISSUE #109: Allow for setting FXA_EMAIL_MESSAGE_ID_HMAC_KEY
We all love ENV vars, but cannot use it for FXA_EMAIL_MESSAGE_ID_HMAC_KEY at this time

 :(

brizental:
Fixes #109 

Because we had underscores in the name of this setting, it wasn't getting overwritten by an `FXA_EMAIL_` environment variable. At first I thought this was a bug with `config-rs` but it actually looks like a design choice, because they use underscores to signal nesting. For example:

```
  "authdb": {
    "baseuri": "http://127.0.0.1:8000/"
  }
```

Will get overwritten by the environment variable `FXA_EMAIL_AUTHDB_BASEURI`.

Anyways, to fix this I changed the `message_id_hmac_key` setting to not have any underscores. Since that was too long for just removing the underscores, I did a little nesting for the resulting environment variable to look nicer. Now, to overwrite the hmac key we can just use `FXA_EMAIL_MESSAGEDATA_HMACKEY`.

I also added that to our settings tests, so that we can make sure it works.

Finally, I remember @philbooth commenting that this is sensitive data that we should probably hide when logging. Should I add a deserializer macro to hide this?

r? @philbooth @vladikoff 

---

[UPDATE]

Looking at `config-rs` source code made me wonder if this really isn't a bug... I filed an issue there (https://github.com/mehcode/config-rs/issues/73) and will try to investigate it.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -19,7 +19,7 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
-  "message_id_hmac_key": "YOU MUST CHANGE ME",
+  "hmackey": "YOU MUST CHANGE ME",
   "logging": "mozlog",
   "provider": "ses",
   "redis": {
--- a/src/message_data/mod.rs
+++ b/src/message_data/mod.rs
@@ -28,7 +28,7 @@ impl MessageData {
             client: RedisClient::open(
                 format!("redis://{}:{}/", settings.redis.host, settings.redis.port).as_str(),
             ).expect("redis connection error"),
-            hmac_key: settings.message_id_hmac_key.clone(),
+            hmac_key: settings.hmackey.clone(),
         }
     }
 
--- a/src/message_data/test.rs
+++ b/src/message_data/test.rs
@@ -63,7 +63,7 @@ impl TestFixture {
     pub fn setup(test: &str) -> TestFixture {
         let settings = Settings::new().expect("config error");
         let unhashed_key = format!("fxa-email-service.test.message-data.{}.{}", test, now());
-        let mut hmac = Hmac::<Sha256>::new_varkey(settings.message_id_hmac_key.as_bytes()).unwrap();
+        let mut hmac = Hmac::<Sha256>::new_varkey(settings.hmackey.as_bytes()).unwrap();
         hmac.input(unhashed_key.as_bytes());
         let internal_key = format!("msg:{:x}", hmac.result().code());
         TestFixture {
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -93,7 +93,7 @@ pub struct Settings {
     pub authdb: AuthDb,
     pub aws: Aws,
     pub bouncelimits: BounceLimits,
-    pub message_id_hmac_key: String,
+    pub hmackey: String,
     pub logging: String,
     #[serde(deserialize_with = "deserialize::provider")]
     pub provider: String,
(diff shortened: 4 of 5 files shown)


config/default.json
```diff
@@ -19,7 +19,9 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
-  "message_id_hmac_key": "YOU MUST CHANGE ME",
+  "messagedata": {
+      "hmackey": "YOU MUST CHANGE ME"
```
philbooth: Fwiw, I maintain my position that we don't need nesting for this and could just have it at the top level (see discussion in https://github.com/mozilla/fxa-email-service/pull/72#discussion_r196182241). Not sure what everyone else thinks though, so ignore me if you guys disagree.

philbooth:
> Looking at `config-rs` source code made me wonder if this really isn't a bug

It's definitely not a bug in `config-rs` fwiw, I think you're supposed to use [`Environment::separator`](https://docs.rs/config/0.8.0/config/struct.Environment.html#method.separator) if you want to work round situations like this. It was just stupid oversight on my part when adding this property, compounded by not including a test for it. So thanks for adding that test here!

config/default.json
philbooth: If we really want to keep the nesting though, maybe it should be inverted in this case. Make `hmackeys` the parent property and then we can have all the different instances that we imagine we want underneath it as children.

brizental: I agree that the nesting is a bit much, but I still think `FXA_EMAIL_MESSAGEDATAHMACKEY` is just too long. I do like the initial solution that you had for this, `FXA_EMAIL_HMACKEY`. I will take a look at the environment separator you mentioned and see if we can keep the underscores after all, if we don't, I will still remove the nesting :)

philbooth: Oh yeah, I definitely didn't mean `FXA_EMAIL_MESSAGEDATAHMACKEY` that's way too ugly. But `FXA_EMAIL_HMACKEY` is fine I think.

philbooth: (but I think @vladikoff disagrees, so let's see what he says)

brizental:
> I think you're supposed to use Environment::separator

So, apparently the `config-rs` crate is an outdated version of it compared to the repo. In the version that we point to (`0.8.0`) the behavior is to use underscore as a separator by default. The repo version doesn't set any default separator anymore (https://github.com/mehcode/config-rs/commit/536f52fed4a22ed158681edce08211845abff985).

This is how this affects us: if we keep using the version of `config-rs` we are using now, when that change comes it will break all our code, because we depend on the default underscore separator to overwrite all our nested settings. What we can do is set the separator even though it is already the default one and when the change comes our code will still work as expected. I think that is a good solution, if we all agree I'll send in a review for this PR.

Regarding having underscores in our settings names, if we are really gonna stick with using underscores as our default separator we can't do that, because whenever we have underscore `config-rs` will always assume that it is something nested inside something else. So, for example, `FXA_EMAIL_MESSAGE_ID_HMAC_KEY` will overwrite this setting:

```
{
   "message": {
      "id": {
         "hmac": {
            "key": "YOU SHOULD CHANGE THIS" # this will be overwritten by FXA_EMAIL_MESSAGE_ID_HMAC_KEY
         }
      }
}
```

and not what we are actually expecting it to overwrite, which is:

```
{
   "message_id_hmac_key": "YOU SHOULD CHANGE THIS" # this won't be overwritten
}
```

This is just something that is nice to know, because I think the solution for the problem that I introduced in this PR (just remove the underscores) is good enough and we shouldn't spend much more time on this. The other solution would be to use another separator to indicate nesting instead of underscores, but that would mean we would have to change ALL our environment variables for settings.

Anyways, let me know what you guys think about this, if we should use another separator or just stick with what we have now.

philbooth:
> Regarding having underscores in our settings names, if we are really gonna stick with using underscores as our default separator we can't do that

Fwiw, my recollection is that we made a conscious decision not to have underscores in setting names back when we did #11, although I can't find a comment about it in that PR so possibly I've misremembered. Anyway, whatever, let's keep on doing that is my thinking. The one you're fixing here was definitely accidental.

config/default.json
vladikoff: I'm sold! I just want my ENV vat :)

config/default.json
```diff
@@ -19,7 +19,7 @@
       { "period": "5 minutes", "limit": 0 }
     ]
   },
-  "message_id_hmac_key": "YOU MUST CHANGE ME",
+  "hmackey": "YOU MUST CHANGE ME",
```
vladikoff: 👍

PR #114, opened 2018-07-03
refactor(settings): refactor settings properties to types

### LINKED ISSUE #25: Refactor settings properties to types
We had some problems with the `config` crate, making it deserialize to our own types. You can see some of the hoops we had to jump through to make that work in the unmerged PR, #14.

At some point we should loop back round to see if we can resolve those issues. It would be nice to decentralise all the `validate` and `deserialize` stuff back to `Deserialize`-implementing type modules and then get rid of those two entirely. We'd also be able to lose the the `deserialize_with = ...` stuff in the `Settings` module.

brizental:
Fixes #25 

I gave this a try with only one of the types in Settings to check that I'm in the right direction.

r? @philbooth


src/settings/mod.rs
```diff
@@ -13,6 +14,11 @@ use serialize;
 #[cfg(test)]
 mod test;
 
+lazy_static! {
+    static ref SQS_URL_FORMAT: Regex =
+        Regex::new("^https://sqs\\.[a-z0-9-]+\\.amazonaws\\.com/[0-9]+/[A-Za-z0-9-]+$").unwrap();
+}
```
philbooth: Let's call the existing function from the `validate` module rather than duplicating the regex inline.

src/validate/mod.rs
```diff
@@ -22,8 +22,6 @@ lazy_static! {
     static ref SENDER_NAME_FORMAT: Regex =
         Regex::new("^[A-Za-z0-9-]+(?: [A-Za-z0-9-]+)*$").unwrap();
     static ref SENDGRID_API_KEY_FORMAT: Regex = Regex::new("^[A-Za-z0-9._-]+$").unwrap();
-    static ref SQS_URL_FORMAT: Regex =
-        Regex::new("^https://sqs\\.[a-z0-9-]+\\.amazonaws\\.com/[0-9]+/[A-Za-z0-9-]+$").unwrap();
```
philbooth: Oh, you were moving it over, not duplicating it, sorry. I still think it's best kept over here though because it makes it much easier to write test cases.

philbooth: (and I quite like how all our validation code lives in the same place, it is especially useful for things like email addresses that are validated both in settings and in request params)

src/settings/mod.rs
```diff
@@ -72,20 +78,34 @@ pub struct Sendgrid {
     pub key: String,
 }
 
+#[derive(Debug, Default, Serialize, PartialEq)]
+pub struct SqsUrl(pub String);
+
+impl<'d> Deserialize<'d> for SqsUrl {
```
philbooth: Actually, there might be an argument for keeping these `impl Deserialize` bits in a dedicated `deserialize` module. Not all of the types we're deserializing are exclusive to `Settings`, so if we want to have the same level of code re-use in the function bodies, it might be easier to put them in one place.

(but also, I might change my mind about that when I see what it actually looks like, not sure yet)

philbooth:
@brizental, you're definitely on the right track!

I'm just wondering about what the right way to split things up is. That might become more obvious if you try to do it for something that's used by both `settings/mod.rs` and `send/mod.rs` (so `email_address`).

Where's the right place for implementing an `EmailAddress` type that's used by both of those?

src/settings/mod.rs
brizental: I actually think it's good to have the `Deserialize` implementation right next to the `struct` definition, makes it easy to find it, if it's standardized that way, all the implementations together with the definition etc.

I think after I do this for more types we will be able to see better if it gets too confusing.

src/validate/mod.rs
brizental: I agree with leaving this stuff inside validation, seems cleaner.

src/settings/mod.rs
brizental: I agree with leaving this stuff inside validation, seems cleaner.

brizental: Yeah, I agree.

src/settings/mod.rs
brizental: Thinking about it now, if that is good it could even become a standard we apply for everything. Always have the `impl` blocks together with the `struct` definitions? Anyways, we'll see, hehe.

philbooth: I agree that the `impl` and `struct` should be next to each other, but I'm not sure that place should be the `settings` module.

brizental:
@philbooth I tried something here. The code was getting very repetitive so I turned the deserializing stuff into a macro. Let me know what you think about that. 

This already changes ALL settings properties that used to have a special serializer/deserializer to types. The only one I left there was `hidden_or_not_set` because that one is different from the other ones and made sense to leave it there.

I know tests are not passing right now, I just wanted to see what were your thoughts on this approach before continuing.

src/bin/queues.rs
```diff
@@ -29,10 +29,10 @@ lazy_static! {
         };
         Queues::new::<Sqs>(
             QueueIds {
-                bounce: sqs_urls.bounce.clone(),
-                complaint: sqs_urls.complaint.clone(),
-                delivery: sqs_urls.delivery.clone(),
-                notification: sqs_urls.notification.clone(),
+                bounce: sqs_urls.bounce.0.clone(),
```
brizental: I don't know if I like adding this `.0` everywhere that we want to access the inner `String` for these new types. Do you have any other solution for this? I could implement `Clone` for the new types easily but that wouldn't fix the whole issue...

src/deserialize/test.rs
```diff
@@ -1,47 +0,0 @@
-// This Source Code Form is subject to the terms of the Mozilla Public
```
brizental: We already test this in the validation tests if I'm not mistaken, so I thought it would be fine to just remove this test.

src/duration/mod.rs
```diff
@@ -56,8 +57,24 @@ impl Display for DurationError {
 /// Can be deserialized from duration strings
 /// of the format `"{number} {period}"`,
 /// e.g. `"1 hour"` or `"10 minutes"`.
-#[derive(Debug, PartialEq)]
-pub struct Duration(u64);
+#[derive(Clone, Debug, Default, Serialize, PartialEq)]
+pub struct Duration(pub u64);
```
brizental: Duration was the only one that was not a string and also the only one that already had a type and didn't use the`validate` module. This is why I didn't use the macro to create it's type.

Also I implemented deserialize here to keep it close to the `struct` declaration.

src/settings/mod.rs
```diff
@@ -1,28 +1,72 @@
...
+            }
+        }
+
+        impl<'d> Deserialize<'d> for $type {
+            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+            where
+                D: Deserializer<'d>,
+            {
+                let value: String = Deserialize::deserialize(deserializer)?;
+                if validate::$validator(&value) {
```
brizental: Maybe here I should be more generic and instead of being bound to the `validate` module I should just receive a `Fn(&str) -> bool` as `$validator` instead...

src/message_data/mod.rs
```diff
@@ -38,7 +38,7 @@ impl MessageData {
     pub fn new(settings: &Settings) -> MessageData {
         MessageData {
             client: RedisClient::open(
-                format!("redis://{}:{}/", settings.redis.host, settings.redis.port).as_str(),
+                format!("redis://{}:{}/", settings.redis.host.0, settings.redis.port).as_str(),
```
brizental: Didn't need the `.0` here...

src/validate/test.rs
```diff
@@ -186,28 +186,29 @@ fn invalid_sendgrid_api_key() {
...
-        " https://sqs.us-east-1.amazonaws.com/123456789012/MyQueue"
-    ));
-    assert!(!validate::sqs_url(
-        "https://sqs.us-east-1.amazonaws.com/123456789012/MyQueue "
-    ));
-    assert!(!validate::sqs_url(
-        "https://sqs.us-east-1.wibble.com/123456789012/MyQueue"
-    ));
-}
+// I know we still need these tests so I'll comment them out for now.
```
brizental: Forgot to un-comment these...

src/bin/queues.rs
philbooth: Although it's a little bit annoying, let's not lose too much sleep over it. Even if we don't come up with something better right now, we might do in the future.

src/settings/mod.rs
```diff
@@ -1,28 +1,72 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
-
```
philbooth: Tiny nit, but can we keep this blank line so that the doc comment is visually separated from the license preamble?

src/settings/mod.rs
```diff
@@ -1,28 +1,72 @@
...
 //! Application settings.
 
-use std::env;
-
 use config::{Config, ConfigError, Environment, File};
+use serde::de::{Deserialize, Deserializer, Error, Unexpected};
+use std::{
+    env,
+    fmt::{self, Display},
+};
```
philbooth: This `use std::...` should stay in its own block above, where it was before.

src/settings/mod.rs
```diff
@@ -1,28 +1,72 @@
...
+    (AwsSecret, aws_secret, "aws secret"),
+    (BaseUri, base_uri, "base uri"),
+    (EmailAddress, email_address, "email address"),
+    (Host, host, "host name or ip address"),
+    (Logging, logging, "'mozlog', 'pretty' ou 'null'"),
+    (Provider, provider, "'ses' or 'sendgrid'"),
+    (SenderName, sender_name, "sender name"),
+    (SendgridApiKey, sendgrid_api_key, "sendgrid api key"),
+    (SqsUrl, sqs_url, "sqs queue url")
+}
```
philbooth: I think this is great!

src/settings/mod.rs
```diff
@@ -1,28 +1,72 @@
...
-use deserialize;
+use duration::Duration;
 use logging::MozlogLogger;
 use serialize;
+use validate;
 
 #[cfg(test)]
 mod test;
 
+macro_rules! deserialize {
```
philbooth: No biggie, but seeing as two of the arguments relate to validation I wonder if the macro should mention validation too? No worries if you think it's too long, but how do you feel about `deserialize_and_validate`?

src/deserialize/test.rs
philbooth: This also asserts that desieralizing the email address lowercases it too. Can we make sure we have that assertion included somewhere else if possible?

philbooth:
> I just wanted to see what were your thoughts on this approach before continuing.

I think it's looking good! With tests it will be perfect! 😄

src/deserialize/test.rs
brizental: I'll look into it, but should be fine :)

src/settings/mod.rs
brizental: Perfect name, agreed.

brizental:
> With tests it will be perfect!

Just leaving this here: tests were already passing I just had forgot to run `fxa-local-dev` while running them................................................................

brizental:
Ended up having to implement deserialize separately for `EmailAddress` because we were making them lowercase on deserialize and it didn't seem right to add a case just for that on the macro. Anyways, final r? @philbooth @vladikoff

src/send/test.rs
```diff
@@ -41,6 +43,46 @@ fn setup() -> Client {
...
+    };
+    assert_de_tokens(
+        &uppercase,
+        &[
+            Token::Struct {
+                name: "TestEmailStruct",
+                len: 1,
+            },
+            Token::Str("email"),
+            Token::Str("FOO@EXAMPLE.COM"),
```
vladikoff: 👍

src/settings/mod.rs
```diff
@@ -189,14 +220,13 @@ pub struct Settings {
...
-    pub logging: String,
+    pub logging: Logging,
 
     /// The default email provider to use,
     /// can be `"ses"`, `"sendgrid"` or `"mock"`.
     /// Note that this setting can be overridden
     /// on a per-request basis.
-    #[serde(deserialize_with = "deserialize::provider")]
-    pub provider: String,
+    pub provider: Provider,
```
vladikoff: the above stuff is cool 👍

src/settings/mod.rs
```diff
@@ -4,25 +4,71 @@
...
+    fmt::{self, Display},
+};
 
 use config::{Config, ConfigError, Environment, File};
+use serde::de::{Deserialize, Deserializer, Error, Unexpected};
 
-use deserialize;
+use duration::Duration;
 use logging::MozlogLogger;
+use send::EmailAddress;
```
philbooth: It's fine for now, no need to change it, but fwiw at some point in the future I might pull `EmailAddress` out to its own module. Something about `settings` depending on `send` feels the wrong way round to me, even though it doesn't actually matter really.

PR #115, opened 2018-07-03
feat(docs): generate developer docs with rustdoc

### LINKED ISSUE #51: cargo doc all the things
Note that cargo doc won't publish comments for non-public types/fields.

philbooth:
Fixes #51.

Adds a bunch of doc comments for use with `rustdoc`. You can try it out locally by running:

```
cargo doc --no-deps --open
```

It should open something like this in your browser:

Screen shot showing the generated docs for fxa_email_service

If this gets a green-light, I'll also create a GitHub Pages branch so that we can refer/link to a hosted version of the docs without needing to build them locally.

@mozilla/fxa-devs r?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,11 +4,11 @@ version = "1.115.0"
 publish = false
 
 [[bin]]
-name = "fxa-email-service"
+name = "fxa_email_send"
 path = "src/bin/service.rs"
 
 [[bin]]
-name = "fxa-email-queues"
+name = "fxa_email_queues"
 path = "src/bin/queues.rs"
 
 [dependencies]
--- a/src/app_errors/mod.rs
+++ b/src/app_errors/mod.rs
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
+//! Error definitions.
+
 use std::{fmt, result};
 
 use failure::{Backtrace, Context, Fail};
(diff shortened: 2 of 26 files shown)


Cargo.toml
```diff
@@ -4,11 +4,11 @@ version = "1.115.0"
 publish = false
 
 [[bin]]
-name = "fxa-email-service"
+name = "fxa_email_send"
```
philbooth: I had to call this something different to prevent a name clash in the generated docs between this bin and the main lib containing all the types etc.

Cargo.toml
```diff
@@ -4,11 +4,11 @@ version = "1.115.0"
...
 path = "src/bin/service.rs"
 
 [[bin]]
-name = "fxa-email-queues"
+name = "fxa_email_queues"
```
philbooth: I swapped hyphens for underscores in both cases because `rustdoc` does that automatically and I figured they should match.

src/app_errors/mod.rs
```diff
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
+//! Error definitions.
```
philbooth: `//!` indicates a module-level comment.

src/app_errors/mod.rs
```diff
@@ -22,6 +24,14 @@ mod test;
 
 pub type AppResult<T> = result::Result<T, AppError>;
 
+/// The main error type
```
philbooth: `///` indicates a type-level comment.

PR #121, opened 2018-07-06
fix(docs): better docs for the settings module

philbooth:
A couple of small documentation fixes, not associated with any issue:

* Add a docstring to the `derive_and_validate` macro expansion. All the other types in `settings` have a doc comment and the generated docs looked incomplete without it for these too.

* Fix up the docs for `SqsUrls`, introducing a newline to keep the outer summary short and moving another note up from the wrong place where it was associated with `SqsUrls::bounce`.

With these changes, the `settings` docs look like this:

Screen shot of the settings documentation

@brizental, @mozilla/fxa-devs r?

### DIFF
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -173,22 +173,23 @@ pub struct Sendgrid {
 }
 
 /// URLs for SQS queues.
+///
 /// Note that these are separate queues right now
 /// for consistency with the auth server.
 /// Long term,
 /// there is nothing preventing us
 /// from handling all incoming notification types
 /// with a single queue.
+///
+/// Queue URLs are specified
+/// for consistency with the auth server.
+/// However, we could also store queue names instead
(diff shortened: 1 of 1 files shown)


src/settings/mod.rs
```diff
@@ -22,7 +22,8 @@ use validate;
 mod test;
 
 macro_rules! deserialize_and_validate {
-    ($(($type:ident, $validator:ident, $expected:expr)),+) => ($(
+    ($(#[$docs:meta] ($type:ident, $validator:ident, $expected:expr)),+) => ($(
+        #[$docs]
```
brizental: Cool stuff, didn't know you had to do this in order to add comments.

philbooth: > didn't know you had to do this in order to add comments

No worries, it took me a while to figure out how to do it! 😄

PR #122, opened 2018-07-10
feat(providers): implement smtp provider

### LINKED ISSUE #46: Implement a local provider that works like mail_helper.js
Ref: https://github.com/mozilla/fxa-auth-server/blob/master/test/mail_helper.js

brizental:
Fixes #46 

I also brought back the SMTP settings :)

r? @philbooth @vladikoff


src/settings/mod.rs
```diff
@@ -172,6 +174,22 @@ pub struct Sendgrid {
     pub key: SendgridApiKey,
 }
 
+/// Settings for SMTP custom provider.
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct Smtp {
+    /// SMTP host IP address.
+    pub host: IpAddress,
```
brizental: Here I'd rather use `std::net::IdAddr` isntead of our own type, but it doesn't implement the `Default` trait and I can't implement it because both the trait and the type are external to this crate. I could remove the `Default` trait from all settings types but that would mean I'd have to change the bounces tests (https://github.com/mozilla/fxa-email-service/blob/master/src/bounces/test.rs#L45). If that is alright with you guys I'll change it.

philbooth: Couldn't this be a host name too? Should we re-use the existing `Host` type?

brizental: That's what I did at first, but here we have to always have either an IPv4 or IPv6 object. A host URI would be invalid. Actually I turn this into a IpAdrr object when I use it.

src/providers/smtp.rs
```diff
@@ -0,0 +1,99 @@
...
+                email = email.header(LettreHeader::new(name.clone(), value.clone()));
+            }
+        }
+
+        if let Some(body_html) = body_html {
+            email = email.html(body_html);
+        }
+
+        let message = email.build()?;
+        let addr = SocketAddr::new(self.host.parse().unwrap(), self.port);
```
brizental: I don't love this unwrap here, but I guess if we change to IpAddrr in the settings we wouldn't need it anyways, so Ill wait.

src/settings/mod.rs
philbooth: Oh, really? Is that because of the lettre API or something?

What would happen if we wanted to [connect to Sendgrid via SMTP](https://sendgrid.com/docs/API_Reference/SMTP_API/getting_started_smtp.html), on `smtp.sendgrid.net`? Or if somebody wanted to connect locally using a host alias from `/etc/hosts` for some reason?

(granted those aren't requirements we have right now, but I'm still curious to understand if/why they're not possible)

philbooth: @brizental, not sure if the example is valid, but for instance [this blog post](http://siciarz.net/24-days-rust-lettre/) shows someone using lettre to send via `smtp.gmail.com`:

```rust
use std::env;

let mut transport = smtp::SmtpTransportBuilder::new(("smtp.gmail.com", smtp::SUBMISSION_PORT))
    .expect("Failed to create transport")
    .credentials(&env::var("GMAIL_USERNAME").unwrap_or("user".to_string())[..],
                    &env::var("GMAIL_PASSWORD").unwrap_or("password".to_string())[..])
    .build();
println!("{:?}", transport.send(email.clone()));
```

pjenvey: You could also impl Default for Smtp manually instead of the #[derive] (it'd be an annoying few lines manually initializing the other fields to Default::default() 🤷)

brizental:
Hey guys, I changed the type of SMTP host to be `Host`. Turns out we can put host names as well as ip addresses there. Also I put credentials inside their own type and then made that type optional, so that when you can't get only username or only password, if you have one you must have both.

About authenticated SMTP calls, I started working on it, but it is a bigger issue and I think it would be best to work on it after this is done. Also, it's not our biggest priority and seems like there are many variables there which make it more complex than it seems. I say we open a follow up issue for that. I also left a `TODO` comment in the code about it.

Anyways, final r? @philbooth @vladikoff

PR #128, opened 2018-07-12
chore(tests): add tests for ses provider and mime message building

### LINKED ISSUE #21: Write tests for the SES provider

brizental:
Fixes #21 

Gets all these done: https://github.com/mozilla/fxa-email-service/issues/21#issuecomment-402312872 

I put the mime message building as a default for the `Provider` trait because it is not something exclusive to the SES provider and might be useful for other providers we might implement in the future and it made it easier for testing.

Also, I removed the underline from all error properties, there was no reason for them to be there and it made it more confusing because it seemed like they had the underline because they were not being used... Let me know if you guys think that should be extracted into it's own PR.

Anyways, r? @philbooth @vladikoff

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -6,6 +6,8 @@
 
 use std::{boxed::Box, collections::HashMap};
 
+use emailmessage::{header, Mailbox, Message, MultiPart, SinglePart};
+
 use self::{
     mock::MockProvider as Mock, sendgrid::SendgridProvider as Sendgrid, ses::SesProvider as Ses,
     smtp::SmtpProvider as Smtp,
@@ -17,6 +19,8 @@ mod mock;
 mod sendgrid;
 mod ses;
 mod smtp;
+#[cfg(test)]
+mod test;
 
 /// Email headers.
 pub type Headers = HashMap<String, String>;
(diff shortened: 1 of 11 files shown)


src/providers/ses/test.rs
```diff
@@ -0,0 +1,63 @@
...
+    };
+    match mock_ses.send("b@b.com", &[], None, "subject", "body", None) {
+        Ok(response) => assert_eq!("woopwoop", response),
+        Err(_) => assert!(false, "Request should have been successfull."),
+    }
+}
+
+#[test]
+fn ses_send_handles_error_response() {
+    let body = r#"FREAKOUT"#;
```
vladikoff: @brizental do we need a test for an SES error structured response? (error xml)

brizental: I'm going to change that line and remove the `r#{...}#` because that is unnecessary.

src/providers/ses/test.rs
```diff
@@ -0,0 +1,63 @@
...
+        client: Box::new(SesClient::new(
+            mock_dispatcher,
+            MockCredentialsProvider,
+            Region::SaEast1,
+        )),
+        sender: "Foo Bar <a@a.com>".to_string(),
+    };
+    match mock_ses.send("b@b.com", &[], None, "subject", "body", None) {
+        Ok(response) => assert_eq!("woopwoop", response),
+        Err(_) => assert!(false, "Request should have been successfull."),
```
philbooth: Fwiw I find it useful, when debugging failing tests, to have the assertion print out the error description instead of a standard message when possible.

src/providers/mod.rs
```diff
@@ -31,6 +35,60 @@ trait Provider {
         body_text: &str,
         body_html: Option<&str>,
     ) -> AppResult<String>;
+
+    fn build_multipart_mime(
```
philbooth: Although I'm 100% in favour of defining this at the generic provider level, I'm not sure about exposing it as part of the `Provider` trait because it's not something we want the calling code to know about. And it doesn't look like `self` is used anyway, so there's no need for it to be a trait method. Could it just be a plain old function and pulled in standalone?

brizental: Done :)

src/providers/mod.rs
```diff
@@ -17,10 +19,66 @@ mod mock;
...
 mod ses;
 mod smtp;
+#[cfg(test)]
+mod test;
 
 /// Email headers.
 pub type Headers = HashMap<String, String>;
 
+/// Build email MIME message.
+pub fn build_multipart_mime(
```
philbooth: Thanks! Let's make it private though, so the tests and the child modules can see it but the send module can't.

PR #134, opened 2018-07-17
feat(project): create healthcheck endpoints and make $PORT an env variable

### LINKED ISSUE #132: implement /__version__, /__heartbeat__, /__lbheartbeat__ routes
Semantics in https://github.com/mozilla-services/Dockerflow#containerized-app-requirements

### LINKED ISSUE #133: Dockerflow apps must listen on environment variable "$PORT"
Is there some other way to set a port to listen on from the environment? Seems to not be on FXA_EMAIL_PORT. The dockerflow standard (and supporting puppet module) want it listen on $PORT.

brizental:
Fixes #132 
Fixes #133 

I have two questions about this:

1. Should I also check if redis is running on `/__heartbeat__`?
2. Should I add tests for each endpoint?

r? @jrgm @vladikoff 

---

[UPDATE]

I added the fix for #133 in this PR as well, now that I look at it maybe I should have created a different one, but let me know what you think.


vladikoff:
1. we can file a follow up issue for redis later
2. tests would be good, other servers test their heartbeat routes

brizental:
> 2. tests would be good, other servers test their heartbeat routes

Alright I'll look at it right now :)

src/settings/mod.rs
```diff
@@ -328,4 +346,26 @@ impl Settings {
...
+                .finalize(),
+            Ok("staging") => RocketConfig::build(RocketEnvironment::Staging)
+                .address(self.host.0.clone())
+                .port(self.port.clone())
+                .log_level(LoggingLevel::Critical)
+                .finalize(),
+            Ok("dev") | _ => RocketConfig::build(RocketEnvironment::Development)
+                .address(self.host.0.clone())
+                .port(self.port.clone())
+                .log_level(LoggingLevel::Normal)
```
brizental: There is a `Debug` level here, but I thought that was a bit too much. This is the default one and the one we were using previously, but just FYI.

Rocket.toml
```diff
@@ -1,12 +0,0 @@
-[development]
```
brizental: This file is never reached if we are going to use `rocket::custom`.

src/healthcheck/test.rs
```diff
@@ -0,0 +1,91 @@
...
+        .mount(
+            "/",
+            routes![super::heartbeat, super::lbheartbeat, super::version],
+        );
+
+    Client::new(server).unwrap()
+}
+
+#[test]
+fn successfull_lbheartbeat() {
```
philbooth: Naming nit: there's only one `l` in `successful` (also below)

src/settings/mod.rs
```diff
@@ -328,4 +337,26 @@ impl Settings {
...
+                .address(self.host.0.clone())
+                .port(self.port.clone())
+                .log_level(LoggingLevel::Off)
+                .finalize(),
+            Ok("staging") => RocketConfig::build(RocketEnvironment::Staging)
+                .address(self.host.0.clone())
+                .port(self.port.clone())
+                .log_level(LoggingLevel::Critical)
+                .finalize(),
+            Ok("dev") | _ => RocketConfig::build(RocketEnvironment::Development)
```
philbooth: Seems like the `| _` makes the `Ok("dev")` redundant here? No biggie of course, just thinking about readability and making the `_` more noticeable if for someone eyeballing the code.

brizental:
Just fixed all the nits, thanks for the review @philbooth 
One thing to note about this PR is that we have no use for `ROCKET_ENV` anymore :)

PR #136, opened 2018-07-19
fix(docker): Rocket.toml no longer exists, so don't copy it

jrgm:
Docker image build fails without removing this line.


jrgm:
@vladikoff - can you tag a v1.116.2 tomorrow as tag v1.116.1 is DOA.

PR #140, opened 2018-07-24
feat(providers): create socketlabs provider

### LINKED ISSUE #123: Prototype SocketLabs provider?
@philbooth should @brizental take a look at how SocketLabs would work with the email service? OR did you have something already in mind / branches?

cc @jbuck

brizental:
Fixes #123 

Before merging I have to add the instructions for this in the `README.md`, but I thought it was better to just send this in for review first. Also before merging, I will send in a new version of the socketlabs-rs to crates.io. We are pointing to the git repo right now because some minor changes came up during implementation and I thought maybe some more necessary changes might come up during review. When the review is done I publish the new version and point to it instead of the repo.

Theres' a lot of `.clone()` going on in the `SocketLabsProvider`. I don't know if that is necessarily a bad thing, but let me know if you guys think of a better way to do it.

Also, SocketLabs apparently doesn't return any identification for the successfully sent messages. It returns a `TransactionReceipt` field for failed requests, but not successful ones. I just put a blank string where `message_id` should go, for now.

r? @philbooth @vladikoff


Cargo.toml
```diff
@@ -48,3 +48,4 @@ slog = ">=2.2.3"
 slog-async = ">=2.3.0"
 slog-mozlog-json = ">=0.1.0"
 slog-term = ">=2.4.0"
+socketlabs = { git = "https://github.com/brizental/socketlabs-rs.git" }
```
vladikoff: can we use a published version ?

src/lib.rs
```diff
@@ -65,19 +65,13 @@ extern crate serde_json;
...
 #[macro_use(
-    slog_b,
-    slog_error,
-    slog_info,
-    slog_kv,
-    slog_log,
-    slog_o,
-    slog_record,
-    slog_record_static
+    slog_b, slog_error, slog_info, slog_kv, slog_log, slog_o, slog_record, slog_record_static
```
vladikoff: were these changes auto formatted?

src/settings/mod.rs
```diff
@@ -289,6 +303,10 @@ pub struct Settings {
 
     /// Settings for SMTP custom provider.
     pub smtp: Smtp,
+
+    /// Settings for SocketLabs.
+    #[serde(serialize_with = "serialize::hidden_or_not_set")]
+    pub socketlabs: Option<SocketLabs>,
```
philbooth: Didn't occur to me last night but should we add a unit test for these settings?

philbooth:
It's probably my fault for doing something wrong with my SocketLabs account set-up, but fwiw I only get error responses when I try to send email:

```
{"code":"500","errno":"104","error":"Internal Server Error","message":"InternalError: Internal server error. (Please report to SocketLabs support if encountered.)","name":"SocketLabs"}
```

No need to block on it working for me though if you guys are both happy.

@brizental, any tips for things I might need to do to get it working?

philbooth:
(`status.socketlabs.com` says everything is operational)

src/settings/mod.rs
brizental: Yessss!

brizental:
@philbooth That is weird, this error is not on the crate nor on email-service. It is a SocketLabs internal error type... They are the ones who return this message. There are very specific messags they return, like if it were message parsing error it would be another message, auth error another message, etc... Anyways, can you try again a little later and see if it works??

src/settings/mod.rs
brizental: Done.

PR #142, opened 2018-07-25
feat(settings): enable the default provider to override requests

philbooth:
Over in https://github.com/mozilla/fxa-auth-server/pull/2535, work is being done to enable the auth server to specify the provider for specific requests. I think this will cause a problem in local dev because we always want to use the local `smtp` provider there, right?

As a hackish approach to resolving that conflict, this change introduces a new boolean setting called `forceprovider`. If `forceprovider` is `true`, the default provider may not be overridden by individual requests.

Maybe there's a less hackish approach though. Any suggestions?

@mozilla/fxa-devs r?

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -26,8 +26,10 @@
     "format": "mozlog"
   },
   "port": 8001,
-  "provider": "ses",
-  "forceprovider": false,
+  "provider": {
+    "default": "ses",
+    "forcedefault": false
+  },
   "redis": {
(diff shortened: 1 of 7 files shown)


philbooth:
> ...changing the config to...

Sure, I can do that!

vladikoff:
@philbooth could you please rebase this?

vladikoff:
Thanks Phil!

PR #147, opened 2018-07-27
feat(settings): add rocket secret_key to settings

### LINKED ISSUE #146: Warning: environment is 'production', but no `secret_key` is configured
one other log message I noticed at startup `Warning: environment is 'production', but no `secret_key` is configured`.

But ROCKET_SECRET_KEY is set in the environment of the docker container process, with a value of $(openssl rand -hex 32)

brizental:
Fixes #146 

When I changed to use `Rocket::custom` instead of `Rocket.toml` for creating configuration, I forgot about `ROCKET_SECRET_KEY`. 

It's good to note that, because we are using `Rocket::custom` none of the `ROCKET_` env vars will work, so if ever we need to set any of the fields in config, we need to add to our config and to our rocket config builder function. 

These are the possible fields in rocket config: https://api.rocket.rs/rocket/config/struct.ConfigBuilder.html#fields

r? @philbooth @jrgm @vladikoff

### DIFF
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -310,6 +310,11 @@ pub struct Settings {
     /// associated with a message.
     pub redis: Redis,
 
+    /// Setting for a secret key
+    /// used by Rocket.
+    #[serde(serialize_with = "serialize::hidden")]
+    pub secretkey: String,
+
     /// Controls the name and email address
     /// that are used for the `From` and `Sender`
     /// email headers.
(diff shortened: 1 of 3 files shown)


src/settings/mod.rs
```diff
@@ -358,8 +363,13 @@ impl Settings {
 
         match config.try_into::<Settings>() {
             Ok(settings) => {
-                if current_env == "production" && &settings.hmackey == "YOU MUST CHANGE ME" {
-                    panic!("Please set a valid HMAC key.")
+                if current_env == "production" {
+                    if &settings.hmackey == "YOU MUST CHANGE ME" {
+                        panic!("Please set a valid HMAC key.")
+                    }
+                    if &settings.secretkey == "youmustchangethisfortheproductionenvironment" {
```
philbooth: Would it make sense to use the same default value for both of these two settings so that we can re-use a constant in these checks? Fwiw `"YOU MUST CHANGE ME"` also appears in the configs for the auth and content servers so seems to be a fairly conventional choice for an FxA repo.

brizental: It would and I did that at first, but this string has to be an exact number of characters or rocket doesn't build... That's I changed it to that weird one, hehe

PR #151, opened 2018-07-30
chore(tests): add tests for request unicode handling

### LINKED ISSUE #9: Ensure unicode handling is sane

brizental:
Fixes #9 

I just added some simple unicode tests for email address, subject and body.

I also added unicode character patterns in our email address regex, because the `.` alone didn't match for some reason...

r? @philbooth @vladikoff

### DIFF
--- a/src/send/test.rs
+++ b/src/send/test.rs
@@ -123,6 +123,82 @@ fn without_optional_data() {
 }
 
 #[test]
+fn unicode_email_field() {
+    let client = setup();
+
+    let mut response = client
+        .post("/send")
+        .header(ContentType::JSON)
+        .body(
+            r#"{
+      "to": "시험@example.com",
+      "subject": "bar",
+      "body": {
+        "text": "baz"
+      },
+      "provider": "mock"
+    }"#,
+        )
+        .dispatch();
+
+    assert_eq!(response.status(), Status::Ok);
+
+    let body = response.body().unwrap().into_string().unwrap();
+    assert_eq!(body, json!({ "messageId": "mock:deadbeef" }).to_string());
+}
+
+#[test]
+fn unicode_message_body() {
+    let client = setup();
+
+    let mut response = client
+        .post("/send")
+        .header(ContentType::JSON)
+        .body(
+            r#"{
+      "to": "foo@example.com",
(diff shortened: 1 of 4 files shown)


src/validate/mod.rs
```diff
@@ -23,21 +23,21 @@ lazy_static! {
     static ref AWS_ACCESS_FORMAT: Regex = Regex::new("^[A-Z0-9]+$").unwrap();
     static ref AWS_SECRET_FORMAT: Regex = Regex::new("^[A-Za-z0-9+/=]+$").unwrap();
     static ref BASE_URI_FORMAT: Regex = Regex::new(
-        "^https?://[A-Za-z0-9-]+(?:\\.[A-Za-z0-9-]+)*(?::[0-9]+)?/(?:[A-Za-z0-9-]+/)*$"
+        r"^https?://[A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)*(?::[0-9]+)?/(?:[A-Za-z0-9-]+/)*$"
```
philbooth: Oh cool, I didn't know regex literals are a thing!

src/send/test.rs
```diff
@@ -122,6 +123,82 @@ fn without_optional_data() {
...
+      "provider": "mock"
+    }"#,
+        )
+        .dispatch();
+
+    assert_eq!(response.status(), Status::Ok);
+
+    let body = response.body().unwrap().into_string().unwrap();
+    assert_eq!(body, json!({ "messageId": "mock:deadbeef" }).to_string());
+}
```
philbooth: These tests are great but is it also possible to add some unicode tests to `src/providers/ses/test.rs`, to make sure that there's no surprises for us at that level?

src/validate/mod.rs
brizental: Yeah, it wouldn't let me do de `\pL` otherwise, said that `\p` wasn't a character they could escape.

brizental: I changed all the other ones as well to avoid that problem and also all the examples in the `Regex` docs are like that (https://docs.rs/regex/1.0.2/regex/)

PR #160, opened 2018-08-07
chore(deps): update rocket, failure, rusoto and rust versions

### LINKED ISSUE #156: Update to failure to 0.1.2 or constrain failure version to 0.1.1
So, I tried updating it today to `0.1.2`, but that returned a bunch of errors that I thought were related to the the `rustc` version we are using, so I updated `rustc` to the newest version. 

That stopped the failure errors but started giving me Rocket errors, which reminded me that that is why we constrained our `rustc` version in the first place: because Rocket needs it.

Long story short, I don't know if we can update failure just yet, because it needs a `rustc` version that Rocket doesn't support, but I might also just be doing something wrong... If that is the case, we should update failure. If that is not the case, we should make sure that failure stays in the `0.1.1` version.

### LINKED ISSUE #158: Update to latest rusoto 0.33
ref https://github.com/rusoto/rusoto/commits/master

brizental:
Fixes #156 
Fixes #158 

Here I update failure, rocket, rusoto and rust to the latest versions.

I had to point rusoto to a revision, because `rusoto_mock` stopped working after the update. I opened an issue about that https://github.com/rusoto/rusoto/issues/1089.

Because I updated the rust version, there are a bunch of formatting changes in this PR, because fmt was updated as well.

r? @philbooth @vladikoff 

_ps.: when this gets merged we can't forget to change the override in `fxa-local-dev` https://github.com/mozilla/fxa-local-dev/blob/master/_scripts/install_all.sh#L42_

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ base64 = ">=0.6.0"
 chrono = { version = ">=0.4.2", features = [ "serde" ] }
 config = ">=0.8.0"
 emailmessage = ">=0.1.0"
-failure = ">=0.1.1"
+failure = ">=0.1.2"
 futures = ">=0.1.21,<0.2"
 hex = ">=0.3.2"
 hmac = ">=0.6.2"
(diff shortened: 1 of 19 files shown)


src/queues/sqs/mod.rs
```diff
@@ -97,9 +97,13 @@ impl Factory for Queue {
         let client: Box<Sqs> = if let Some(ref keys) = settings.aws.keys {
             let creds =
                 StaticProvider::new(keys.access.0.clone(), keys.secret.0.clone(), None, None);
-            Box::new(SqsClient::new(RequestDispatcher::default(), creds, region))
+            Box::new(SqsClient::new_with(
+                HttpClient::new().expect("Couldn't start HTTP Client."),
```
brizental: I'm not really sure about this `expect` here, what do you think @philbooth ? Should I change this functions to return a `Result` and handle the errors or panicking is alright in this case?

philbooth: I think panicking in this case is fine because it's initialisation code, so the process will abort on startup rather than later when we think everything is working correctly.

Cargo.toml
```diff
@@ -30,14 +30,14 @@ redis = ">=0.8.0"
...
-rusoto_ses = ">=0.32.0"
-rusoto_sqs = ">=0.32.0"
+rocket = { git = "https://github.com/SergioBenitez/Rocket", rev="d7f6d82fe4d35ee79c424dd4164733b22f17f80a" }
+rocket_codegen = { git = "https://github.com/SergioBenitez/Rocket", rev="d7f6d82fe4d35ee79c424dd4164733b22f17f80a" }
+rocket_contrib = { git = "https://github.com/SergioBenitez/Rocket", rev="d7f6d82fe4d35ee79c424dd4164733b22f17f80a" }
+rusoto_core = { git = "https://github.com/rusoto/rusoto/", rev="9e22eb3876a02375b3dbe14d0340f9b5c1fc55e9" }
+rusoto_credential = { git = "https://github.com/rusoto/rusoto/", rev="9e22eb3876a02375b3dbe14d0340f9b5c1fc55e9" }
+rusoto_mock = { git = "https://github.com/rusoto/rusoto/", rev="9e22eb3876a02375b3dbe14d0340f9b5c1fc55e9" }
+rusoto_ses = { git = "https://github.com/rusoto/rusoto/", rev="9e22eb3876a02375b3dbe14d0340f9b5c1fc55e9" }
+rusoto_sqs = { git = "https://github.com/rusoto/rusoto/", rev="9e22eb3876a02375b3dbe14d0340f9b5c1fc55e9" }
```
philbooth: It mildly concerns me that our dependence on git hashes instead of released versions of crates appears to be growing rather than shrinking. Are we really so desperate for the targeted features (including for the preceding one that caused the initial Rocket dependency on a commit hash) that we can't wait for them to show up in official releases?

brizental: Yes, I don't like that either... in this case we are pointing to the 0.33.0 release commit, so it should be fine. Also I will change this to an actual version as soon as https://github.com/rusoto/rusoto/issues/1089 gets done.

We probably could have waited for them to release the rusoto version with the fixes that we need for `rusoto_mock`, but since we wanna test ASAP if the new release fixes our SES problems (#159) I just pointed to the revision for now.

The rocket revision concerns me more because that is almost a random commit, but apparently they are going to release the new version before RustConf or during RustConf, anyways, we will be able to change that soon :)

brizental:
Thanks, @philbooth ! I made the docker changes and created a PR in local-dev to update rust version :)

vladikoff:
> I am a little bit uncomfortable that dependening on git hashes is becoming more prevalent, but I'll defer to your and @vladikoff's judgement on that point.

Yeah agreed, we can file follow up issues to get rid of `git` dependencies. Rusoto is a special one though because we are tracking multiple issues with it...

Issue #116, opened 2018-07-03
Improve logging and error messages


brizental:
This needs a little bit more digging, but there are some log/error messages that could be improved.

Like, if we try and send in a request with missing params, it will return the following log:

```
{"Logger":"fxa_email_service-1.115.0","Type":"fxa_email_service:log","Pid":91132,"Severity":3,"Timestamp":1530658150217212000,"Fields":{"msg":"{\"code\":\"400\",\"errno\":\"101\",\"error\":\"Bad Request\",\"message\":\"Missing email params.\"}","agent":"curl/7.54.0","remoteAddressChain":"127.0.0.1","path":"/send","method":"POST"}}
```

Here, it would be nicer to know, for example, which are the missing params. In auth server we just add more properties to some of our error messages when they need more info (https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-format), that could be a solution here.

Anyways, probably not something urgent, but I just noticed this and thought would be best to document this here.

philbooth:
Ah, yeah, there used to be a `TODO` comment about this but I can't find it now, so thanks for raising this!

Anyway, for the validation errors, I think it depends on a resolution to SergioBenitez/Rocket#596 happening first.

PR #161, opened 2018-08-08
chore(queues): use mozlog and failure in queues process

### LINKED ISSUE #10: Implement logging

### LINKED ISSUE #99: Improve error handling in the 'queue' process
Remaining errors that need to use `failure`:

- [ ] messagedata
- [ ] duration

Ref: https://github.com/mozilla/fxa-email-service/pull/94

brizental:
Fixes #10 
Fixes #99 
Connects to #116 

This was my first experience messing with the queues part of the service, I thought dealing with these issues first would make the experience smoother. I added the standard logging and error handling with the failure crate just like the rest of the lib. 

Since our errors were very much centered around requests and rocket, we have some fields that don't really make sense for the queues process, like the http code, so let me know what you think about that and if the queues process should have a different error format. I also did another minor change to the errors: now the response JSON will return a number for `code` and `errno`, instead of a string.

For logging, I implemented an `AppErrorFields` so that we get better structured logs for our `AppError`s. I used `slog_scope` to make it easier to have a global logger for the queues process, let me know what you think about that... In the `README.md` for that crate, they have some warnings about it not being the best idea all the time, but anyways, I thought it was neat and worked well for our case.

Finally, when I started working in the queues they were not really working due to parsing errors, you will see that I changed a little bit the SQS Notification struct, that was for parsing to work, also I created a notification-dev queue in the AWS Console, because it didn't exist yet, I think everything is working fine now.

r? @philbooth @vladikoff


src/app_errors/mod.rs
```diff
@@ -45,15 +45,18 @@ impl AppError {
...
         fields.insert(
             String::from("error"),
             Value::String(format!("{}", status.reason)),
         );
         let errno = kind.errno();
         if let Some(ref errno) = errno {
-            fields.insert(String::from("errno"), Value::String(format!("{}", errno)));
+            fields.insert(
+                String::from("errno"),
+                Value::Number(errno.to_owned().into()),
```
philbooth: Ref: https://github.com/mozilla/fxa-auth-server/issues/2569

src/app_errors/mod.rs
```diff
@@ -145,9 +148,47 @@ pub enum AppErrorKind {
...
-    #[fail(display = "{:?}", _0)]
+    #[fail(display = "{}", _0)]
     DbError(String),
 
+    /// An error for when an error happens on the queues process.
+    #[fail(display = "{}", _0)]
+    QueueError(String),
+    /// An error for when we get an invalid notification type in the queues
+    /// process.
+    #[fail(display = "Invalid notification type.")]
```
philbooth: Tiny nit: I'm not a fan of including a terminating period in our error strings fwiw. None of our other repos do it and it may be incorrect in the context of the receiving code, if it wants to display the message as part of a broader error string.

src/logging/mod.rs
```diff
@@ -59,6 +60,46 @@ impl KV for RequestMozlogFields {
     }
 }
 
+#[derive(Clone)]
+struct AppErrorFields {
+    code: String,
```
philbooth: Doesn't really matter tbh, it's only logging, but this one could/should be an `i32` as well, right?

philbooth: (actually `u16` is probably more correct for both if I'm feeling pedantic, which I usually am)

src/queues/sqs/notification/mod.rs
```diff
@@ -139,10 +140,10 @@ pub struct Mail {
     sending_account_id: Option<String>,
     destination: Option<Vec<String>>,
     #[serde(rename = "headersTruncated")]
-    headers_truncated: Option<String>,
+    headers_truncated: Option<bool>,
     headers: Option<Vec<Header>>,
     #[serde(rename = "commonHeaders")]
-    common_headers: Option<Vec<Header>>,
+    common_headers: Option<HashMap<String, HeaderValue>>,
```
philbooth: Should the `Header` circa line 160 be deleted too?

src/queues/sqs/notification/mod.rs
```diff
@@ -137,10 +140,10 @@ pub struct Mail {
     sending_account_id: Option<String>,
     destination: Option<Vec<String>>,
     #[serde(rename = "headersTruncated")]
-    headers_truncated: Option<String>,
+    headers_truncated: Option<bool>,
```
philbooth: Ref: https://github.com/mozilla/fxa-email-service/issues/162

src/app_errors/mod.rs
```diff
@@ -250,12 +331,17 @@ impl From<Context<AppErrorKind>> for AppError {
 /// Generate HTTP error responses for AppErrors
 impl<'r> Responder<'r> for AppError {
     fn respond_to(self, request: &Request) -> response::Result<'r> {
+        match request.guard::<State<MozlogLogger>>() {
+            Outcome::Success(logger) => {
+                let log = MozlogLogger::with_app_error(&logger, &self)
+                    .map_err(|_| Status::InternalServerError)?;
+                slog_error!(log, "{}", "Request errored.");
+            }
+            _ => println!("Internal error: No managed MozlogLogger"),
```
philbooth: How come this isn't a  `slog_error!`?

philbooth: Duh, because haven't got a logger you div.

In that case, should we panic here? If logging isn't working we probably want to abort don't we?

src/queues/sqs/notification/mod.rs
philbooth: Oh, wait, no we shouldn't, it's still used for the `headers` property on the line above.

(apologies for my poor reading comprehension! 😄)

brizental: I thought about writing a deserializing function for the `common_headers` so that we can keep it as a `Option<Vec<Header>>`, it makes sense to me that both headers are the same type... But then I thought that maybe we might want the struct to mimic exactly the message we get. What do you think, @philbooth ?

src/app_errors/mod.rs
brizental: Yes, panicking sounds reasonable.

src/queues/sqs/notification/mod.rs
philbooth: I reckon keep it like you've got here, it's probably less confusing if we match the SES format as closely as possible.

philbooth:
I'm happy, we can merge this as soon as the conflict is resolved!

brizental:
Would you like me to squash these commits before merging also, @philbooth ?

Issue #139, opened 2018-07-24
requests to SES API mysteriously stop working, but few details on why...


jrgm:
/cc @brizental, @philbooth, @vladikoff 

So, what had been working for fxa-email-service, stopped working today. I thought maybe I had done something dumb, and redeployed in stage, and it started working again. Ah, so I **am** dumb!

And then with no changes, it stopped working. Ok, not me being dumb, for now.

I did an `strace(1)` to confirm that the error occurs on the call to SES, and not a problem with setting up that request. (`strace` checks with auth-db (OK), gets credentials (OK), negotiates TLS connection to SES (OK), so not a problem in setup).

The error either comes from SES, or in fxa-email-service handling whatever response came back. (Note: the connection to SES is TLS, so `strace` cannot show me that response in raw HTTP).

Unfortuately, the failing error message (with NODE_ENV='not-production') doesn't provide much detail on root SES error reason. (Note: `sendto(10` is the HTTP response back to the auth-server; `write(1` is logging to stdout).

`
[pid  4570] 02:59:53.631182 sendto(10, "HTTP/1.1 500 Internal Server Error\r\nConnection: close\r\nContent-Type: application/json\r\nServer: Rocket\r\nContent-Length: 99\r\nDate: Tue, 24 Jul 2018 02:59:53 GMT\r\n\r\n{\"code\":\"500\",\"errno\":\"104\",\"error\":\"Internal Server Error\",\"message\":\"Unknown(\\\"\\\")\",\"name\":\""..., 261, MSG_NOSIGNAL, NULL, 0 <unfinished ...>

[pid  4547] 02:59:53.631202 write(1, "{\"Logger\":\"fxa_email_service-1.116.2\",\"Type\":\"fxa_email_service:log\",\"Pid\":1,\"Severity\":3,\"Timestamp\":1532401193631141906,\"Fields\":{\"msg\":\"{\\\"code\\\":\\\"500\\\",\\\"errno\\\":\\\"104\\\",\\\"error\\\":\\\"Internal Server Error\\\",\\\"message\\\":\\\"Unknown(\\\\\\\"\\\\\\\")\\\",\\\"name\\\":\\\""..., 330 <unfinished ...>
`

We're going to need a way to enable more detailed logging of rust code and dependencies to get further on this. (I can file a ticket with with SES, but the first thing they'll want are those details on the HTTP response from SES).

jrgm:
500/104 is ProviderError, ya?

[...]

PR #163, opened 2018-08-09
fix(ses): use rusoto version that sends message in body

brizental:
Connects to #139 

This fixes the issue, but I didn't mark it as fixed because right now we are pointing to my fork.

You can see the whole discussion here: https://github.com/rusoto/rusoto/issues/1093 and this is the easiest way to test it: https://github.com/mozilla/fxa-email-service/issues/139#issuecomment-411584944

This PR is just so that we can test stuff :)


vladikoff:
@jrgm i am closing this PR , but you can still use the `feature.test-ses` image to deploy this

PR #172, opened 2018-08-15
feat(queues): add hyper logs for queues

brizental:
cc @vladikoff 

run this like this -> `RUST_LOG=rusoto,hyper=debug  cargo run --bin fxa_email_queues`

PR #174, opened 2018-08-15
fix(queues): fix executor error on queues process

brizental:
See discussion #172 

@seanmonstar thanks for telling us how to fix this :)

r? @vladikoff @philbooth


src/lib.rs
```diff
@@ -81,6 +81,7 @@ extern crate slog_mozlog_json;
 extern crate slog_scope;
 extern crate slog_term;
 extern crate socketlabs;
+extern crate tokio;
```
philbooth: Does it need to be added to the lib too? Looks like only bin/queues references it?

vladikoff: yes @brizental !

brizental: Done!

PR #176, opened 2018-08-16
feat(logging): add sentry support

### LINKED ISSUE #155: Integrate Sentry into this project
autopush-rs has this already  https://github.com/mozilla-services/autopush-rs/blob/ef59d10ae32982811da7f6f799fe4a2bbbfc6e62/Cargo.toml#L47 but several versions behind

Ref: https://crates.io/crates/sentry

brizental:
Fix #155 

I added support for Sentry. For that to work we just need to set `FXA_EMAIL_SENTRY_DSN`.

Right now we are logging to Sentry everytime we log an `AppError`.

r? @vladikoff

### DIFF
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -77,6 +77,8 @@ deserialize_and_validate! {
     (SenderName, sender_name, "sender name"),
     /// Sendgrid API key type.
     (SendgridApiKey, sendgrid_api_key, "Sendgrid API key"),
+    /// Sentry DSN type.
+    (SentryDsn, sentry_dsn, "sentry dsn"),
     /// AWS SQS queue URL type.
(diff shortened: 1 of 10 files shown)


src/settings/mod.rs
```diff
@@ -205,6 +207,13 @@ pub struct Sendgrid {
     pub key: SendgridApiKey,
 }
 
+/// Settings for Sentry.
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct Sentry {
+    /// The Sentry Data Source Name
+    pub dsn: SentryDsn,
```
philbooth: Should we have two separate DSN's, one for the service and another for the queues?

vladikoff: @philbooth probably overkill

PR #177, opened 2018-08-17
fix(logging): remove unnecessary and add debug logs in queues

brizental:
Fixes #173

There is a bunch of minor fixes going on in this PR.

- Fix #173 
- Add some debug logs for notifications in case we need it (probably would be good to only show them when we have some `debug` flag on or something)
- Notification JSON may come in different shapes, nested or not nested, so I added functionality to accept both. That logic might me a little fragile so let me know what you think.

r? @vladikoff @philbooth

PR #178, opened 2018-08-21
fix(queues): make queue parsing logic follow the sns format

brizental:
This is just a follow up PR from #177 

I remove the part where I make the parser work for different SNS message formats, turns out that is not needed. 

See discussion: https://github.com/mozilla/fxa-email-service/pull/177#issuecomment-414394797

r? @vladikoff

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,7 +67,6 @@ extern crate serde_test;
 extern crate sha2;
 #[macro_use(
     slog_b,
-    slog_debug,
     slog_error,
     slog_info,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -67,7 +67,6 @@ extern crate serde_test;
 extern crate sha2;
 #[macro_use(
     slog_b,
-    slog_debug,
```
brizental: I removed this because there was a warning about this not being used.

PR #180, opened 2018-08-22
feat(errors): return bouncedAt field in bounce error message

### LINKED ISSUE #157: Include a timestamp in bounce/complaint responses
The auth server includes a `bouncedAt` property on the bounce/complaint errors. We should return that from the email service so it can be set correctly.

brizental:
Fixes #157 

r? @philbooth @vladikoff

### DIFF
--- a/src/app_errors/mod.rs
+++ b/src/app_errors/mod.rs
@@ -134,17 +134,20 @@ pub enum AppErrorKind {
     #[fail(display = "Email account sent complaint")]
     BounceComplaintError {
         address: String,
-        bounce: Option<BounceRecord>,
+        bounced_at: u64,
+        bounce: BounceRecord,
     },
     #[fail(display = "Email account soft bounced")]
     BounceSoftError {
         address: String,
-        bounce: Option<BounceRecord>,
+        bounced_at: u64,
+        bounce: BounceRecord,
     },
     #[fail(display = "Email account hard bounced")]
     BounceHardError {
         address: String,
-        bounce: Option<BounceRecord>,
+        bounced_at: u64,
+        bounce: BounceRecord,
     },
 
     /// An error for when an error happens on a request to the db.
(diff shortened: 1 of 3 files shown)


src/app_errors/mod.rs
```diff
@@ -134,16 +135,19 @@ pub enum AppErrorKind {
     #[fail(display = "Email account sent complaint")]
     BounceComplaintError {
         address: String,
+        bounced_at: DateTime<Utc>,
```
philbooth: Is it okay if we add a `#[serde(rename = "notificationType")]` attribute to these, so that the serialized property name matches what's expected by the auth server?

https://github.com/mozilla/fxa-auth-server/blob/3d5fd4c1bcd495b767c77dfac83a0d4a93994519/lib/senders/email_service.js#L57

philbooth: Whoops, copy-paste fail, I meant `bouncedAt` not `notificationType` of course. 😊

philbooth: Also, is `DateTime` serialised as a string or as an epoch timestamp? The auth server tests look like they're expecting the latter, just wanted to check we're doing the same thing here:

https://github.com/mozilla/fxa-auth-server/blob/3d5fd4c1bcd495b767c77dfac83a0d4a93994519/test/local/senders/index.js#L293

src/bounces/mod.rs
```diff
@@ -70,14 +72,17 @@ where
                         return match bounce.bounce_type {
                             DbBounceType::Hard => Err(AppErrorKind::BounceHardError {
                                 address: address.to_string(),
+                                bounced_at: Utc::now(),
```
philbooth: I think, rather than setting these to `now`, maybe they should have the same value as `bounce.created_at` instead? Semantically, in the error message we're really saying this is when the bounce occurred, rather than this is when we noticed it in the db.

(in which case my question about `DateTime` serialization above is moot anyway I guess, since `created_at` is a u64 so we could just propagate that directly)

philbooth: Arrgh, sorry for all the comments on this one! Can we also add a couple of tests?

Either in the bounces tests or the app_error ones, wherever you think it makes most sense, just to assert that `bouncedAt` is serialized and has a value that matches the bounce record?

philbooth: (I guess that means in the bounce tests, because they have stuff for creating bounce records)

philbooth: (and you probably don't even need a new test case there, just an assertion in an existing one)

src/app_errors/mod.rs
brizental: > Is it okay if we add a #[serde(rename = "notificationType")] attribute to these, so that the serialized property name matches what's expected by the auth server?

I think that is not necessary, since we are not implementing Serialize at all for this enum. The way we get the data we need is through the `additional_fields` function.

PR #184, opened 2018-09-17
chore: utilize a rust-toolchain file

pjenvey:
moves the toolchain spec to solely one file


vladikoff:
would the rustup override from the readme override this? do we need to remove that?

philbooth:
> do we need to remove that?

Oh yeah, well spotted! (and actually that readme was out of date anyway since we pinned to 2018-08-06).

vladikoff:
Yeah, the readme was the other file we always had to bump the version in

Issue #162, opened 2018-08-08
Test message-parsing in the queues process


philbooth:
When we added email headers in #89, notification-parsing in the queues process stopped working. It stopped working because the `Notification` struct was wrong, but we didn't notice because there were no automated tests for the parsing code. When I manually tested it while implementing, everything seemed fine because there were no headers.

The fix for this particular parsing issue is included in #161:

https://github.com/mozilla/fxa-email-service/pull/161/files#diff-d73554e88b8870cc1336cbfba3460e2eR143

But we still don't have any tests to ensure it won't break again in the future. We should add some proper parsing tests, using real data copied from manual testing.

philbooth:
Added this to the Q4 milestone.

brizental:
This was fixed by #188

PR #188, opened 2018-09-21
chore(tests): add tests for message parsing in the queues

brizental:
Connects to #162 

I started working on this issue, but I'm not really sure I'm doing the right thing. I'm submitting this work in progress so that you can check that this is the kind of tests needed, @philbooth . I started the code for the tests, but before finishing it I just wanted to check, because it looks kind of weird to me...

I also added a commit that updates the config.rs version in the Cargo.toml because we had `0.8.0` there, but turns out we use features from the crate that don't exist in that version only on `>=0.9.0`. That was a problem I stumbled on while working on #162 this is why the commit is in this PR. It was such a small change I didn't think it needed to have a whole PR for itself, but let me know if you want I can do that too.

r? @philbooth

### DIFF
--- a/src/queues/sqs/notification/test.rs
+++ b/src/queues/sqs/notification/test.rs
@@ -243,3 +243,92 @@ fn serialize_complaint_feedback_type() {
     let json = serde_json::to_string(&ComplaintFeedbackType::Virus).expect("JSON error");
     assert_eq!(json, "\"virus\"");
 }
+
+#[test]
+fn deserialize_notification() {
+    let json = include_str!("test_notifications.json");
+    let notifications: Vec<Notification> = serde_json::from_str(json).expect("JSON error");
+
+    assert_eq!(notifications.len(), 3);
+
+    assert_eq!(notifications[0].notification_type, NotificationType::Bounce);
+    assert_eq!(
+        notifications[0].mail.source,
+        Some("foo@example.com".to_string())
+    );
+    assert_eq!(
+        notifications[0].mail.destination,
+        Some(["bar@example.com".to_string()].to_vec())
+    );
+    assert!(notifications[0].bounce.is_some());
+    assert!(notifications[0].complaint.is_none());
+    assert!(notifications[0].delivery.is_none());
+    let bounce = match notifications[0].bounce {
+        Some(ref record) => record,
+        _ => panic!("Bounce record should exist."),
+    };
+    assert_eq!(bounce.bounced_recipients.len(), 3);
+    assert_eq!(
+        bounce.bounced_recipients[0].email_address,
+        "bar@example.com".to_string()
+    );
+    assert_eq!(
+        bounce.bounced_recipients[1].email_address,
+        "baz@example.com".to_string()
+    );
+    assert_eq!(
+        bounce.bounced_recipients[2].email_address,
+        "qux@example.com".to_string()
+    );
+    assert_eq!(bounce.bounce_type, BounceType::Permanent);
+    assert_eq!(bounce.bounce_subtype, BounceSubtype::General);
+
+    assert_eq!(
+        notifications[1].notification_type,
+        NotificationType::Complaint
+    );
+    assert_eq!(
+        notifications[1].mail.source,
+        Some("qux@example.com".to_string())
+    );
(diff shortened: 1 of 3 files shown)


src/queues/sqs/notification/test.rs
```diff
@@ -243,3 +244,127 @@ fn serialize_complaint_feedback_type() {
...
+            },
+            "delivery": {
+                "timestamp":"2014-05-28T22:41:01.184Z",
+                "processingTimeMillis":546,
+                "recipients":["xyzzy@example.com"],
+                "smtpResponse":"250 ok:  Message 64111812 accepted",
+                "reportingMTA":"a8-70.smtp-out.amazonses.com",
+                "remoteMtaIp":"127.0.2.0"
+            }
+        }]
```
philbooth: I read about a cool trick the other day that may or may not be handy here: [include_str!](https://doc.rust-lang.org/std/macro.include_str.html)

Basically we could put this big blob of JSON into a file in the same directory and call it something like `test_data.json` (or whatever you think makes sense), then `include_str!` reads it into a `&str` for us in one step.

Just keeps this file a bit smaller and makes things like editing easier, because you're editing a `.json` file so editors and IDEs know how to behave etc.

src/queues/sqs/notification/test.rs
```diff
@@ -243,3 +244,127 @@ fn serialize_complaint_feedback_type() {
...
+    assert_eq!(notifications[0].notification_type, NotificationType::Bounce);
+    assert_eq!(notifications[0].mail.timestamp, Utc.ymd(2016, 01, 27).and_hms(14, 05, 45));
+    assert_eq!(notifications[0].mail.message_id, "001");
+    assert_eq!(notifications[0].mail.source, Some(String::from("foo@example.com")));
+    assert_eq!(notifications[0].mail.source_arn, Some(String::from("arn:aws:ses:us-west-2:888888888888:identity/example.com")));
+    assert_eq!(notifications[0].mail.source_ip, Some(String::from("127.0.1.0")));
+    assert_eq!(notifications[0].mail.sending_account_id, Some(String::from("1234")));
+    assert_eq!(notifications[0].mail.destination, Some([String::from("bar@example.com")].to_vec()));
+    assert_eq!(notifications[0].mail.headers_truncated, Some(false));
+    // ...
```
philbooth: No need to assert every single property I think, just `notification_type` and any properties that are inserted as part of the bounce record should be sufficient. And an assertion for `notifications.len()`.

Actually, a bit stupidly on my part, I only just understood a couple of weeks ago that I didn't need to create all these properties when I was defining the `Notification` struct. For some reason, back then I assumed serde would fail to deserialize if I hadn't defined every single property. But of course, it doesn't, it just skips over any that aren't defined.

At some point we can go through and remove most of those properties because they're just getting in the way, making the code longer, and we're never going to use them.

philbooth: Ref: #189

brizental: > No need to assert every single property I think

Perfect! That was my confusion :)

> At some point we can go through and remove most of those properties because they're just getting in the way, making the code longer, and we're never going to use them.

Great!

src/queues/sqs/notification/test.rs
```diff
@@ -243,3 +243,47 @@ fn serialize_complaint_feedback_type() {
...
+
+    assert_eq!(notifications[0].notification_type, NotificationType::Bounce);
+    assert!(notifications[0].bounce.is_some());
+    assert!(notifications[0].complaint.is_none());
+    assert!(notifications[0].delivery.is_none());
+    let bounce = match notifications[0].bounce {
+        Some(ref record) => record,
+        _ => panic!("Bounce record should exist."),
+    };
+    assert_eq!(bounce.bounced_recipients.len(), 3);
```
philbooth: Hey @brizental, one last thing we should assert is that the email addresses in here match the test data. Then we're golden!

brizental: Done!

philbooth:
Thanks @brizental, I'm really happy to have this stuff covered by tests!

PR #190, opened 2018-09-25
fix(queues): remove unused fields from notification structs

### LINKED ISSUE #189: Rip out all the cruft from the Notification structure
For some reason, when I wrote the `Notification` struct, I assumed serde would fail to deserialize if I hadn't defined every single property that AWS sets on the object. But of course it doesn't, it just skips over any that aren't defined on the target object.

We should remove all the unused properties from our `Notification` struct and make that module much, much shorter.

brizental:
Fixes #189 

Removed everything that is not used anywhere else in the code but the `id`s because I thought that we might need them if ever we need to debug something.

r? @philbooth

### DIFF
--- a/src/queues/sqs/notification/mod.rs
+++ b/src/queues/sqs/notification/mod.rs
@@ -4,10 +4,7 @@
 
 //! SQS queue notification types.
 
-use std::{
-    collections::HashMap,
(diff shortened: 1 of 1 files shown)


philbooth:
Thanks @brizental!

PR #192, opened 2018-09-26
fix(config): ignore empty environment variables

### LINKED ISSUE #191: Allow "empty" settings in environment variables
In the fxa nightly stacks, these values were set in the environment (i.e., "empty" strings).

```
FXA_EMAIL_SENTRY_DSN=
FXA_EMAIL_SOCKETLABS_SERVERID=
FXA_EMAIL_SOCKETLABS_KEY=
```

These cause fxa-email-service to crash with errors such as 
invalid type: string "", expected an integer in the environment
or
invalid value: string "", expected sentry dsn

IMHO, in other server programs, these "empty" environment values would be treated as unset and a program default value would be used.

It's possible to configure `nightly` differently. However, having a different semantics for empty environment variables from what other stacks would have will be a pain.

philbooth:
Fixes #191.

You can also test it manually by setting, say, `FXA_EMAIL_PORT` to the empty string and running `./r`. Before this change that fails with the error:

```
thread 'main' panicked at 'Settings::new error: invalid type: string "", expected an integer in the environment', libcore/result.rs:983:5
```

@mozilla/fxa-devs r?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,7 +177,7 @@ dependencies = [
 
 [[package]]
 name = "config"
-version = "0.9.0"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "lazy_static 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


philbooth:
Thanks @shane-tomlinson!

vladikoff:
👍

Issue #166, opened 2018-08-11
Migrate away from using fxa-auth-db's database & schema


jbuck:
There's two parts to this:

1) From an operational POV, once we're only using fxa-email-service in production we can spin up a new database rather than reusing the fxa-auth-db's database.
2) It'd be good to add database migrations to this repo at that point in time

philbooth:
Moved this into the milestone I'm using for Q4 work.

jbuck:
Redis, MySQL, and PostgreSQL are all supported on GCP. I'd recommend simplicity and choose one data store for all types of data. Redis feels like the right choice to me given the requirements, though it does mean we'd need to migrate data from MySQL to Redis. Not the end of the world, but worth noting.

philbooth:
I'm definitely happy with Redis, although I've noticed that when I suggest it for FxA stuff people often pull a funny face at me and say something like "we need a proper database for this" <sup>[citation needed]</sup>.

So, just checking (and I don't really know what I'm talking about), are we happy with the consistency / persistence guarantees with Redis? I guess bounce records and metrics data are both ephemeral to some extent anyway so it's not the end of the world if we were to lose some, but the configuration data not so much.

Also, just linking to earlier discussion about this, in #4 @rfk cautioned that some of the data will be sensitive and that we should keep that in mind when picking a data store.

jbuck:
If the auth server needs to store relational data than MySQL or Postgres totally make sense, but from your description of the data being stored, it sounds like Redis is a good fit. Maybe post your ideal structs for each of the data sets (bounce records, config, metrics) in this bug and we can decide for sure?

I'm happy with the consistency guarantees of Redis, as it's a single-threaded service so everything is atomic by default. The persistence guarantees I'm happy enough so long as we use a highly-available setup in prod, which we do for all of our Redis instances so far.

In terms of sensitive data we follow the principle of least-access so only the service, @jrgm, and I should have access

jbuck:
If the auth server needs to store relational data than MySQL or Postgres totally make sense, but from your description of the data being stored, it sounds like Redis is a good fit. Maybe post your ideal structs for each of the data sets (bounce records, config, metrics) in this bug and we can decide for sure?

I'm happy with the consistency guarantees of Redis, as it's a single-threaded service so everything is atomic by default. The persistence guarantees I'm happy enough so long as we use a highly-available setup in prod, which we do for all of our Redis instances so far.

[...]

PR #193, opened 2018-10-01
refactor(redis): extract redis code to a dedicated db module

philbooth:
Related to #166.

We've decided to use Redis as the data store for all types in this repo so as a first step towards that, this PR just pulls the redis code out of `message_data` into its own module.

No functionality has changed so no tests are changed. And no new tests are added because the `db` module is going to be tested via its consumers (i.e. the existing `message_data` tests right now, but also `bounces` and `configuration` in the future).

@mozilla/fxa-devs r?

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1,0 +1,113 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Database abstractions.
+//!
+//! Uses Redis under the hood,
+//! which is fine because
+//! none of the data is relational
+//! and we flip all of the
+//! Redis persistence switches in prod.
+//! You can read more about this decision
+//! in [#166](https://github.com/mozilla/fxa-email-service/issues/166).
+
+use std::fmt::{self, Display, Formatter};
+
+use hmac::{crypto_mac::InvalidKeyLength, Hmac, Mac};
+use redis::{Client as RedisClient, Commands, RedisError};
+use sha2::Sha256;
+
+use app_errors::{AppError, AppErrorKind, AppResult};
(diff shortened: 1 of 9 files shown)


src/healthcheck/mod.rs
```diff
@@ -2,10 +2,11 @@
...
 
-//! Route handlers for our heathcheck endpoints:
-//! for the `GET /__version__` endpoint,
-//! for the `GET /__lbheartbeat__` endpoint and
-//! for the `GET /__heartbeat__` endpoint,
+//! Route handlers for our heathcheck endpoints.
+//!
+//! * `GET /__version__`
+//! * `GET /__lbheartbeat__`
+//! * `GET /__heartbeat__`
```
philbooth: Opted to tweak this while I was in here, because it looked a bit busy in the generated docs with them all on the same line:

Screenshot of the generated documentation for the healthcheck module

src/db.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl From<RedisError> for AppError {
+    fn from(error: RedisError) -> AppError {
+        AppErrorKind::DbError(format!("redis error: {:?}", error)).into()
+    }
+}
+
+impl From<InvalidKeyLength> for AppError {
+    fn from(error: InvalidKeyLength) -> AppError {
+        AppErrorKind::DbError(format!("hmac key error: {:?}", error)).into()
```
brizental: I wonder if we should create a specific error for this one, like `AppErrorKind::HmacKeyError`, and remove the prefixes (`redis error: ` and `hmac key error: `) from these two. The advantage of that is a clearer error message and a specific errno for each. 

This is not necessarily something I think should be changed, when creating the errors for this app I was never sure about the best way to do this, what do you think, @philbooth ?

philbooth: > I was never sure about the best way to do this

Me neither!

> what do you think, @philbooth ?

I'll make the change!

PR #196, opened 2018-10-04
refactor(types): prefer EmailAddress type to raw strings

philbooth:
The EmailAddress type was only being used in some places, which meant some places that weren't using it had to call `validate::email_address` manually. This change spreads it throughout most of the project, so that almost everywhere gets to benefit from strong typing.

The only place I decided to leave alone was the provider layer, where some providers have their own `EmailAddress` struct. I could have aliased it at that layer too, but by that point addresses are just dumb strings anyway so it didn't seem worth it.

(this is an extraction from `pb/166-wip` because that branch is growing too big)

@mozilla/fxa-devs r?

### DIFF
--- a/src/auth_db/test.rs
+++ b/src/auth_db/test.rs
@@ -138,18 +138,8 @@ fn serialize_bounce_subtype() {
 fn get_bounces() {
     let settings = Settings::new().expect("config error");
     let db = DbClient::new(&settings);
-    if let Err(error) = db.get_bounces("foo@example.com") {
+    if let Err(error) = db.get_bounces(&"foo@example.com".parse().unwrap()) {
         assert!(false, format!("{}", error));
-    }
-}
-
-#[test]
-fn get_bounces_invalid_address() {
-    let settings = Settings::new().expect("config error");
-    let db = DbClient::new(&settings);
-    match db.get_bounces("") {
-        Ok(_) => assert!(false, "DbClient::get_bounces should have failed"),
-        Err(error) => assert_eq!(format!("{}", error), "400 Bad Request"),
     }
 }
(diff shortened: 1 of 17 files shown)


src/auth_db/test.rs
```diff
@@ -138,21 +138,11 @@ fn serialize_bounce_subtype() {
...
 
-#[test]
-fn get_bounces_invalid_address() {
-    let settings = Settings::new().expect("config error");
-    let db = DbClient::new(&settings);
-    match db.get_bounces("") {
-        Ok(_) => assert!(false, "DbClient::get_bounces should have failed"),
-        Err(error) => assert_eq!(format!("{}", error), "400 Bad Request"),
-    }
-}
```
philbooth: Here we can delete the entire test case because it's not possible to construct an invalid `EmailAddress` instance to pass to `db.get_bounces`.

src/send/mod.rs
```diff
@@ -64,24 +56,6 @@ impl FromData for Email {
...
 }
 
-fn validate(email: &Email) -> bool {
-    if let Some(ref cc) = email.cc {
-        for address in cc {
-            if !validate::email_address(&address) {
-                return false;
-            }
-        }
-    }
```
philbooth: Again, all this ugly stuff can go now because the validation is already encapsulated inside `EmailAddress`.

philbooth:
> ...someone with more rust chops should have a look too.

Merging anyway, because I have a queue of PRs that are partially blocked by this one and it's not clear if anyone else is going to review it or not.

If anyone does have further feedback on this PR, feel free to mention it here and I can address it separately.

PR #198, opened 2018-10-05
refactor(bounces): pull all bounce/complaint code into one module

philbooth:
Related to #166. Depends on #196.

This change does two things:

1. The old `BounceRecord`, `BounceType` and `BounceSubtype` types are removed from the `auth_db` module into `bounces`, so that they can more easily be re-used when writing bounce/complaint data to Redis.

2. Those types are renamed to `DeliveryProblem`, `ProblemType` and `ProblemSubtype` respectively, and the `bounces` module is renamed to `delivery_problems`. That may seem weird at first, but I have good reason for it and am open to alternative names if anyone can suggest something better.

   Historically we've always treated complaints as a type of bounce event, but that's inaccurate because bounces are never preceded by a delivery event whereas complaints are always preceded by one. This has caught me out in the past when analysing metrics and expecting a blanket sum of `deliveries + bounces = sends` to be true. It isn't. It's only true if you filter complaints from bounces before calculating the sum.

   Because of that, I wanted some different nomenclature that didn't use "bounces" as an umbrella term for "bounces and complaints". It won't affect the names we use in the metrics but I think it's important for language we use in the codebase to be as precise as possible. Things like "delivery error" or "delivery failure" also seemed inaccurate for the same reason; a complaint implies that the actual delivery succeeded. Hence "delivery problems", which seems generic enough to legitimately include both event types. But I'm not wedded to that particular name so feel free to suggest something different.

Marking as `WIP` for now (even though I'm happy with it) because it's based off the branch for #196 and will need rebasing against `master` after that lands (this is another extraction from the `pb/166-wip` branch).


philbooth:
> ...it's based off the branch for #196 and will need rebasing against master after that lands...

Rebased (and ready for review).

src/app_errors/mod.rs
```diff
@@ -133,22 +133,22 @@ pub enum AppErrorKind {
 
     /// An error for when a bounce violation happens.
     #[fail(display = "Email account sent complaint")]
-    BounceComplaintError {
+    ComplaintError {
         address: EmailAddress,
         bounced_at: u64,
-        bounce: BounceRecord,
+        bounce: DeliveryProblem,
```
philbooth: I haven't changed these property names because it makes the handling code less uniform. I figure consumers will have an easier time if they're consistent with the bounce errors. I could easily be persuaded otherwise, though.

philbooth: (we should make our minds up about it soon though, because this is part of the published API)

rfk:
A huge +1 to the nomenclature adjustments here FWIW

API.md
```diff
@@ -45,9 +45,9 @@ The following errors include additional response properties:
 - `errno: 103`
 - `errno: 104` name
 - `errno: 105`
-- `errno: 106` address, bouncedAt, bouce
-- `errno: 107` address, bouncedAt, bouce
-- `errno: 108` address, bouncedAt, bouce
+- `errno: 106` address, bouncedAt, bounce
```
rfk: It seems unfortunately that the "Complaint Error" still has a "bouncedAt" field, but perhaps we can't help that for API-compatibility reasons.

src/app_errors/mod.rs
rfk: Could they be kept uniform by using a more generic term like "deliveryProblemType" and "deliveryProblemOccurredAt" or similar, in line with your renaming of the module?

src/auth_db/mod.rs
```diff
@@ -11,205 +11,32 @@
...
-    }
-
-    pub fn create_bounce(&self) -> Url {
-        self.create_bounce.clone()
-    }
-}
-
 pub trait Db: Debug + Sync {
-    fn get_bounces(&self, address: &EmailAddress) -> AppResult<Vec<BounceRecord>>;
+    fn get_bounces(&self, address: &EmailAddress) -> AppResult<Vec<DeliveryProblem>>;
```
rfk: Should these method names also change from "bounce" to "delivery problem"?

src/delivery_problems/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+        let bounce_subtype = complaint_type.map_or(ProblemSubtype::Unmapped, |ct| From::from(ct));
+        self.auth_db
+            .create_bounce(address, ProblemType::Complaint, bounce_subtype)?;
+        Ok(())
+    }
+}
+
+unsafe impl<D> Sync for DeliveryProblems<D> where D: AuthDb {}
+
+fn is_bounce_violation(count: u8, created_at: u64, now: u64, limits: &[BounceLimit]) -> bool {
```
rfk: This could be `is_limit_violation` or similar, to lean into the terminology change.

src/auth_db/mod.rs
philbooth: My thinking for these was twofold:

1. That they should reflect the naming in the auth db, since they're a an abstraction thereof.
2. They kind of don't matter as much as the rest of the code because we'll be deleting this `auth_db` module eventually.

But those are not strongly-held opinions, so I'll make the change.

rfk: My opinion is also not strongly held FWIW, so I'll defer to your judgement either way :-)

philbooth:
Further changes have been pushed, including to the error structure. Since that is part of the API already used by the auth server, this PR now depends on mozilla/fxa-auth-server#2663 being merged first.

src/auth_db/mod.rs
philbooth: I decided against this one in the end. I think it's right that the names here match the names in the db repo.

src/delivery_problems/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+use queues::notification::{BounceSubtype, BounceType, ComplaintFeedbackType};
+use settings::{BounceLimit, BounceLimits, Settings};
+
+#[cfg(test)]
+mod test;
+
+/// Bounce/complaint registry.
+///
+/// Currently this writes to
+/// both Redis and
```
rfk: It doesn't appear to write to Redis, is this comment out-of-date?

philbooth: Oh whoops, it's an extraction from the branch that *does* write to Redis, which is my next PR. So, yes, out of date but not in the conventional direction. I'll remove it from this diff.

PR #200, opened 2018-10-05
feat(db): serialize to JSON on write and deserialize from JSON on read

philbooth:
Related to #166.

Previously, our Redis abstraction only worked with strings because it only dealt with the somewhat nebulous concept of message metadata. Now that we also want to use it for concrete types, it makes sense to bake in automatic serialization from and deserialization to those types. This frees consumers from the responsibility of having to serialize to a Redis-friendly format manually.

This change achieves that aim by making the db methods generic. `db.set` gets a type parameter that is `Serialize` and the getters get one that is `DeserializeOwned`.

(this is the final extraction from `pb/166-wip` before I can open the actual pull request for that change)

@mozilla/fxa-devs r?

### DIFF
--- a/src/app_errors/mod.rs
+++ b/src/app_errors/mod.rs
@@ -201,6 +201,10 @@ pub enum AppErrorKind {
     /// An error occured while hashing a value.
     #[fail(display = "HMAC error: {}", _0)]
     HmacError(String),
+
+    /// An error occured while serializing or deserializing JSON.
+    #[fail(display = "JSON error: {}", _0)]
+    JsonError(String),
 }
 
 impl AppErrorKind {
@@ -251,6 +255,8 @@ impl AppErrorKind {
             AppErrorKind::NotImplemented => Some(118),
 
             AppErrorKind::HmacError(_) => Some(119),
+
+            AppErrorKind::JsonError(_) => Some(120),
 
             AppErrorKind::BadRequest
             | AppErrorKind::NotFound
--- a/src/message_data/test.rs
+++ b/src/message_data/test.rs
@@ -31,7 +31,7 @@ fn set() {
             .unwrap();
         assert!(!key_exists, "unhashed key should not exist in redis");
         let value: String = test.redis_client.get(test.internal_key.as_str()).unwrap();
-        assert_eq!(value, "wibble");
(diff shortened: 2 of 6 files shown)


src/queues/sqs/mod.rs
```diff
@@ -222,6 +222,6 @@ impl From<DeleteMessageError> for AppError {
 
 impl From<JsonError> for AppError {
     fn from(error: JsonError) -> AppError {
-        AppErrorKind::QueueError(format!("JSON error: {:?}", error)).into()
+        AppErrorKind::JsonError(format!("JSON error: {:?}", error)).into()
```
philbooth: Changed this to something more broadly applicable so that I could re-use it in the `db` module.

src/app_errors/mod.rs
```diff
@@ -201,6 +201,10 @@ pub enum AppErrorKind {
     /// An error occured while hashing a value.
     #[fail(display = "HMAC error: {}", _0)]
     HmacError(String),
+
+    /// An error occured while serializing or deserializing JSON.
```
rfk: What sort of errors might occur during serialization?

rfk: Serde says "Serialization can fail if T's implementation of Serialize decides to fail, or if T contains a map with non-string keys." so fair enough...

src/message_data/test.rs
```diff
@@ -31,7 +31,7 @@ fn set() {
             .unwrap();
         assert!(!key_exists, "unhashed key should not exist in redis");
         let value: String = test.redis_client.get(test.internal_key.as_str()).unwrap();
-        assert_eq!(value, "wibble");
+        assert_eq!(value, "\"wibble\"");
```
rfk: So will this change the serialization of existing datatypes, meaning an existing db will now contain invalid data?  Or were all uses of this already storing a JSON string in practice?

philbooth: Strictly speaking it's a breaking change. Data already stored would now fail to be read.

But actually there's no data in there at the moment because the auth server isn't posting any to `/send` yet. And even if there was, we could probably still get away with this because the reads are only in the queues process right now, which we still haven't enabled (so we could just drop any data if there was some).

PR #202, opened 2018-10-15
refactor(redis): stop treating no data as an error in the getters

philbooth:
This is another extraction from the branch for #166.

Previously, we were treating an absence of data as an error case in Redis. That's wrong for every data type we plan to store, so this PR changes the `db` module to wrap its result types in an `Option`.

@mozilla/fxa-devs r?

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -47,7 +47,7 @@ impl Client {
     }
 
     /// Read data.
-    pub fn get<D>(&self, key: &str, data_type: DataType) -> AppResult<D>
+    pub fn get<D>(&self, key: &str, data_type: DataType) -> AppResult<Option<D>>
     where
         D: DeserializeOwned,
     {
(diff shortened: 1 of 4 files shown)


src/db.rs
```diff
@@ -47,19 +47,23 @@ impl Client {
...
+    pub fn get<D>(&self, key: &str, data_type: DataType) -> AppResult<Option<D>>
     where
         D: DeserializeOwned,
     {
         let key = self.generate_key(key, data_type)?;
         self.client
             .get(key.as_str())
             .map_err(From::from)
-            .and_then(|value: String| serde_json::from_str(&value).map_err(From::from))
+            .and_then(|value: Option<String>| {
```
vladikoff: 👍

src/db.rs
```diff
@@ -68,9 +72,11 @@ impl Client {
         self.client
             .get(key_str)
             .map_err(From::from)
-            .and_then(|value: String| {
-                self.client.del::<&str, u8>(key_str).ok();
-                serde_json::from_str(&value).map_err(From::from)
+            .and_then(|value: Option<String>| {
```
vladikoff: 👍

PR #203, opened 2018-10-16
feat(db): write bounce and complaint records to our own db

philbooth:
Related to #166.

This is the first step towards migrating away from the FxA auth db. We'll start by dual-writing to both databases. Then, after we've got sufficient history to fully evaluate limit violations using our own data, we can switch over to read from our own datastore and ditch the auth db.

Note that we still need to migrate from the auth server's bounce and complaint handling logic in prod before this code starts doing anything.

@mozilla/fxa-devs r?

### DIFF
--- a/src/delivery_problems/mod.rs
+++ b/src/delivery_problems/mod.rs
@@ -77,17 +77,17 @@ where
                             ProblemType::HardBounce => Err(AppErrorKind::BounceHardError {
                                 address: address.clone(),
                                 time: problem.created_at,
-                                problem: problem.clone(),
+                                problem: From::from(problem.clone()),
                             }.into()),
                             ProblemType::SoftBounce => Err(AppErrorKind::BounceSoftError {
                                 address: address.clone(),
                                 time: problem.created_at,
-                                problem: problem.clone(),
+                                problem: From::from(problem.clone()),
                             }.into()),
                             ProblemType::Complaint => Err(AppErrorKind::ComplaintError {
                                 address: address.clone(),
                                 time: problem.created_at,
-                                problem: problem.clone(),
+                                problem: From::from(problem.clone()),
                             }.into()),
                         };
                     }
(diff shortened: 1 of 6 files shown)


src/delivery_problems/mod.rs
```diff
@@ -101,8 +105,35 @@ where
...
+        address: &EmailAddress,
+        problem_type: ProblemType,
+        problem_subtype: ProblemSubtype,
+    ) -> AppResult<()> {
+        let mut problems: Vec<DeliveryProblem> = self
+            .db
+            .get(address.as_ref(), DataType::DeliveryProblem)?
+            .unwrap_or_else(|| Vec::new());
+        problems.insert(
+            0,
```
rfk: IIUC, inserting at the head of a `Vec` like this going to be O(N) in the number of delivery problems in the list, because it has to shift every other element to the right.  Is it possible to append to the end and then traverse the vector in reverse order, or something like that?

What will happen if we have so many bounce records for an address that the serialized value exceeds the maximum size of a string in redis?  Should we purge older entries to keep the size of the list under control?

We expect the size of the list to be small in practice, but we don't want to open any DoS vectors here.

(I know it would be a much bigger PR, but part of me wonders whether we should use redis's builtin list datatype for this...)

philbooth: My plan was to purge older items, as per the discussion at https://github.com/mozilla/fxa-email-service/issues/166#issuecomment-425035671. But I can look at appending and then changing the logic that processes the list instead, no probs.

Althoug also, just fwiw, it wouldn't be possible to DoS the service from this code as it is not invoked from the API. We'd DoS the queues process instead, meaning we'd be slow to process bounce/complaint/delivery notification.

rfk: Actually, is it true that we want the items in this list to be sorted by timestamp?  Inserting at the head or tail of the list won't necessarily guarantee that, if the SQS processing happens to deal with messages out of order (which it can).  Perhaps we should explicitly insert in sorted order here to be sure.

rfk: Although I guess we just generate the timestamp ourselves don't we, rather than taking a timestamp from the message....

I should probably stop worrying at this point and just leave it in your capable hands :-)

philbooth: Yeah, timestamps are generated in the auth db at the moment, presumably because we decided it was risky to trust the time reported by SES. I'm in favour of revisiting that decision as part of this broader migration process fwiw, we're already trusting SES to report accurate bounce/complaint events to us so trusting the timestamp too doesn't seem too big a leap.

philbooth:
@brizental, @mozilla/fxa-devs I pushed a couple more commits to this branch so it probably needs a fresh r+ before it can be merged, sorry.

* In 1093258, I make the change suggested by @rfk in https://github.com/mozilla/fxa-email-service/pull/203#discussion_r225678922, to store items in the opposite order to the auth db so that we can append to the array rather than shift every item to the right. Note that because we don't read from this data store, it's a smaller change, just the `append` and a big fat `TODO` comment for when we do read the items from Redis.

* In b2773f5, I opt to better separate the old and new naming schemes by not separating our `DeliveryProblem` from the old `LegacyDeliveryProblem`. The properties are exactly the same, just the serialization attributes are changed. It means we won't end up with the bounce umbrella terminology leaking into our data store and error structures. I should have thought of this one originally, sorry.

r?

src/delivery_problems/mod.rs
philbooth: > I'm in favour of revisiting that...

Ref: #205

philbooth:
@vbudhram yeah, we're still pinned to a really old version of Rust, sorry. If you do:

```
rustup toolchain install nightly-2018-08-06
rustup override set nightly-2018-08-06
```

...it might work?

We need to update some deps and then update the Rust version to something more recent, it's been on my to-do list for a while!

philbooth:
Sorry, ignore that comment, just seen you're on the correct version in your comment for #206.

PR #208, opened 2018-10-29
fix(queues): fix the serialized format of outgoing notifications

philbooth:
The `bounced_recipients` property of the outgoing notification structure was missing a serde `rename` attribute. That problem went undetected because we had no test coverage for the outgoing structure, which was a pretty large hole in our coverage. This change fixes the issue and adds some serialization assertions.

@mozilla/fxa-devs r?


philbooth:
(also, as per the discussion in retro, I'm happy to go through this in Vidyo if it's helpful for potential reviewers)

PR #209, opened 2018-10-29
chore(deps): update versions of Rust, Rocket and Rusoto

### LINKED ISSUE #207: Update deps and Rust version
We've been pinned to nightly-2018-08-06 for ages, because of some deps that we have pinned to git hashes that won't build with more recent versions. We should check whether the commits we pinned for have now been released more generally and then update to a more recent Rust version if so.

philbooth:
Fixes #207.

We've been pinned to an ancient version of Rust for a while and I really wanted to get us updated before cutting train 124 (which is hopefully going to see some traffic). Updating Rust required updating Rocket and Rusoto, so that's done here too. Initially I wanted to do a full `cargo update` but the build fails if you do that because of rust-lang/rust#55376. As soon as that's resolved, we should update the other deps too.

Most of the code changes here are because `cargo fmt`. There's also a couple of knock-on code changes from the Rocket update (the disappearance of `rocket_contrib::JsonValue` and the disappearance of `rocket::LoggingLevel::Off`).

Because there's a bit going on, I've kept everything as separate commits. If we want to review them separately I'm happy to break this into multiple PRs. Also, as discussed in retro recently, if you want to join me in Vidyo so I can take you through the changes in more detail, I am more than happy to do that.

@mozilla/fxa-devs r?


src/bin/service.rs
```diff
@@ -58,7 +51,7 @@ fn main() {
     let config = settings
         .build_rocket_config()
         .expect("Error creating rocket config");
-    rocket::custom(config)
+    rocket::custom(config, settings.log.level.0.as_str() != "off")
```
rfk: Is the `.0` here going to conflict with you other PR where you remove the use of this `.0` style?

src/lib.rs
```diff
@@ -24,7 +24,6 @@
 //! [send]: ../fxa_email_send/index.html
 //! [queues]: ../fxa_email_queues/index.html
 
-#![feature(assoc_unix_epoch)]
```
rfk: What's this one from?

Issue #216, opened 2018-10-31
Convert string settings types to enums where possible


philbooth:
The following types all have a limited set of values and should probably modelled as enums rather than strings:

* [ ] ~~`settings::AwsRegion`~~
* [x] `settings::DefaultProvider`
* [x] `settings::Env`
* [x] `settings::LoggingLevel`
* [x] `settings::LoggingFormat`

philbooth:
> `settings::DefaultProvider`

Ref: #218

vladikoff:
from mtg: @philbooth still 126?

philbooth:
> * [ ] `settings::AwsRegion`

Correction, let's just use [`rusoto_core::Region`](https://rusoto.github.io/rusoto/rusoto_core/region/enum.Region.html) for that one instead.

PR #218, opened 2018-10-31
feat(settings): extract provider type to a fully-fledged enum

philbooth:
Related to #216.

The provider type was modelled as a newtype struct around a `String` but there's a finite number of values, so really it should be an enum. This wasn't too pressing until now but will shortly come in useful for the configuration stuff, which needs a provider type too.

@mozilla/fxa-devs r?

### DIFF
--- a/src/settings/mod.rs
+++ b/src/settings/mod.rs
@@ -21,7 +21,9 @@ use rocket::config::{
 use serde::de::{Deserialize, Deserializer, Error, Unexpected};
 
 use logging::MozlogLogger;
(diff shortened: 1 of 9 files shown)


src/settings/mod.rs
```diff
@@ -67,8 +69,6 @@ deserialize_and_validate! {
     (AwsSecret, aws_secret, "AWS secret key"),
```
vbudhram: You mentioned some follow up work here to expose these as enums. Should we make a bug to track that?

philbooth: Way ahead of you: #216 :smile:

PR #228, opened 2018-11-07
fix(email): fix sentry errors in prod

### LINKED ISSUE #220: Sentry errors in SesProvider::send
Our first Sentry error for this repo! :tada:

https://sentry.prod.mozaws.net/operations/email-service/issues/4900545/?environment=release

Backtrace:

```
    <unknown>0x55f030cf6792<backtrace::capture::Backtrace as core::default::Default>::default
    Called from: backtracebacktrace::capture::Backtrace::new
    <unknown>0x55f0307f8c37<std::thread::local::LocalKey<T>>::with
    Called from: sentrysentry::hub::Hub::with_active
    <unknown>0x55f030dfa265rust_begin_unwind panicking.rs:325
    Called from: corecore::panicking::panic_fmt panicking.rs:77
    <unknown>0x55f030e485d6<core::result::Result<T, E>>::expect result.rs:809
    Called from: allocalloc::fmt::format fmt.rs:570
    <unknown>0x55f0306f9cee<fxa_email_service::providers::ses::SesProvider as fxa_email_service::providers::Provider>::send
    fxa_email_service0x55f0306dcd00fxa_email_service::providers::Providers::send
    fxa_email_service0x55f03075d60bfxa_email_service::send::handler
    fxa_email_service0x55f03075d0e9fxa_email_service::send::rocket_route_fn_handler
    rocket0x55f030b40b50rocket::rocket::Rocket::route_and_process
    <unknown>0x55f030b3ec88<rocket::rocket::Rocket as hyper::server::Handler>::handle
    <unknown>0x55f030b2e90a<hyper::server::Worker<H>>::handle_connection
    hyper0x55f030b45fa7hyper::server::listener::spawn_with::{{closure}}
    Called from: stdstd::sys_common::backtrace::__rust_begin_short_backtrace
    <unknown>0x55f030b1d8d2<F as alloc::boxed::FnBox<A>>::call_box
    <unknown>0x55f030dfc48a<alloc::boxed::Box<(dyn alloc::boxed::FnBox<A, Output=R> + 'a)> as core::ops::function::FnOnce<A>>::call_once boxed.rs:652
    Called from: stdstd::sys_common::thread::start_thread thread.rs:24
    <unknown>0x7fa09b6bb493start_thread
    <unknown>0x7fa09b1e6ace__clone
    <unknown>0x000000000000<unknown> 
```

philbooth:
Fixes #220 (maybe, hopefully).

There are actually three changes in here:

* In 85ffe3e, I update the `emailmessage` dependency. I don't *think* this impacts either Sentry error but after I opened an issue there was a flurry of activity in the repo, so I opted to include those changes.

  Note that the API changed fairly significantly, so most of the diff comes from this commit. Perhaps most importantly, we lost the ability to serialize arbitrary headers because of the interface provided between the `Header` trait and the `Raw` struct. The only way I could see to make it work was to implement `Header` ourselves for each header value that we expect, hence the new module in `src/types/headers`.

* In 5966d62, I elminated the `format!` invocations in the SES provider. Again I don't *think* this fixes the root cause, but one of the panics was occurring inside a `format!` so I did it anyway because I couldn't repro said panic locally.

* In 5d03a48, I finally fix what I *think* is the root cause of both errors, which is that we were sending 8-bit encoded MIME data to SES, whereas the [docs](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-raw.html#send-email-mime-encoding-body) suggest that we should only send quoted-printable or base-64:

  > #### Message Body
  >
  > To encode the body of a message, you can use quoted-printable encoding or base64 encoding.

   I think that's where the `parsed HTTP message from remote is incomplete` error stems from.

  For the panic, although I couldn't reproduce it locally, I did see strange behaviour where the 8-bit encoded message sometimes caused the process to hang when I tried to serialize it to a string. I think that may be a different manifestation of the same problem.

  The `emailmessage` crate docs aren't *massively* detailed and the [RFC](https://tools.ietf.org/html/rfc6152) text is pretty dense so it's likely our expectations about the 8-bit encoding type were wrong in some way. But whatever, it seems academic now because SES doesn't want us to use 8-bit encoding and base-64 encoding makes my hang go away, so let's just use it and see where it gets us.

Opened against `train-124` so that we can tag a patch release and see what effect it has. Also because that may help us get the OKR back on track, if we're able to start sending some traffic through the email service without waiting for train 125 to roll out.

@mozilla/fxa-devs r?


src/providers/test.rs
```diff
@@ -71,9 +72,9 @@ fn build_mime_with_custom_headers() {
     assert_eq!("To: b@b.com", &message[1]);
     assert_eq!("Subject: subject", &message[2]);
     assert_eq!("MIME-Version: 1.0", &message[3]);
-    assert_eq!("x-foo: bar", &message[4]);
+    assert_eq!("X-Device-Id: baz", &message[4]);
```
vladikoff: @philbooth this new test update for header capital case, could this break some other tests where we have something like `header['x-link']` instead of `X-Link`

src/providers/mod.rs
```diff
@@ -27,57 +30,73 @@ mod test;
...
+        );
     }
 
-    Ok(message.with_body(MultiPart::mixed().with_multipart(body)))
+    Ok(message.mime_body(MultiPart::mixed().multipart(body)))
+}
+
+fn set_custom_header(message: MessageBuilder, name: &str, value: &str) -> MessageBuilder {
+    let lowercase_name = name.to_lowercase();
+    match lowercase_name.as_str() {
```
vladikoff: 👍

Cargo.toml
```diff
@@ -15,11 +15,12 @@ path = "src/bin/queues.rs"
 base64 = ">=0.6.0"
 chrono = { version = ">=0.4.2", features = [ "serde" ] }
 config = ">=0.9.1"
-emailmessage = ">=0.1.0"
+emailmessage = ">=0.2.0"
```
vladikoff: 👍

src/providers/test.rs
vladikoff: ok sounds good, there might be other external tests that might be affected (fxa pom), we just need to remember that this was the change.

PR #230, opened 2018-11-07
chore(logging): updates for sentry

### LINKED ISSUE #222: Include debug symbols for more detailed Sentry errors
Noticed [this](https://blog.sentry.io/2018/05/01/sentry-for-rust):

> Errors will contain as much information as possible at event submission, so the most important step is to make sure to ship your debug information in the executable. Experimental support for server side symbolication is also provided if you upload debug symbols.

Currently we're not including debug symbols in the release build. We can do that by setting `debug = true` in the `[profile.release]` section of Cargo.toml.

philbooth:
Fixes #222.

* In 4636112, I update the Sentry dependency to the latest version.

* In bfeb745, I move the point at which we capture errors for Sentry, so that we now catch every error rather than just those which are logged.

* In 5faef07, I turn on debug symbols in the release build, so that Sentry errors contain richer information.

@mozilla/fxa-devs r?


Cargo.toml
```diff
@@ -11,6 +11,10 @@ path = "src/bin/service.rs"
 name = "fxa_email_queues"
 path = "src/bin/queues.rs"
 
+[profile.release]
+# Include debug symbols to make Sentry errors richer
+debug = true
```
vladikoff: nice! I thought it said this auto 👍

philbooth:
> In 5faef07, I turn on debug symbols in the release build, so that Sentry errors contain richer information.

Just noting that this had the desired effect, we now get line numbers in the stack traces.

Issue #185, opened 2018-09-18
Add a configuration endpoint


philbooth:
Essentially we want to move functionality from the auth server's live email config, added in mozilla/fxa-auth-server#2574, to here. I think the best way of doing that is to add a new endpoint that accepts configuration.

Later, when we're opening this service up to more clients, we can ensure that the configs are identified by client so that the correct config is applied to each `/send` request.

This depends on the data store that we're going to add for #166.

vladikoff:
from mtg: @philbooth still 126?

PR #231, opened 2018-11-07
feat(bounces): use timestamps from ses instead of the current time

### LINKED ISSUE #205: Use the provider timestamp when storing bounce and complaint records in our own db
The auth db elected to use the current time when storing bounce records, for reasons that are unclear. We don't need to make the same decision here though, and I am in favour of using the provider-supplied timestamp instead so that bounce/complaint records are a more accurate representation of when the problem occurred rather than when we happened to process the event.

philbooth:
Fixes #205.

SES provides a UTC-based timestamp for bounce and complaint events but the auth server opted to ignore it. I'm not sure that was the optimal decision because messages might linger on a queue (either ours or the provider's) for a length of time before we process it. Given that we already trust SES et al to deliver sane notifications, it doesn't seem too great a stretch to also trust the timestamp on those notifications.

@mozilla/fxa-devs r?

### DIFF
--- a/src/db/delivery_problems/mod.rs
+++ b/src/db/delivery_problems/mod.rs
@@ -9,6 +9,7 @@ mod test;
 
 use std::{collections::HashMap, time::SystemTime};
 
+use chrono::{DateTime, TimeZone, Utc};
 use serde::{
     de::{Deserialize, Deserializer, Error as DeserializeError, Unexpected},
     ser::{Serialize, Serializer},
@@ -64,6 +65,7 @@ where
     /// [limits]: ../settings/struct.BounceLimits.html
     pub fn check(&self, address: &EmailAddress) -> AppResult<()> {
         let problems = self.auth_db.get_bounces(address)?;
+        // TODO: When we start reading from the new datastore, use Utc::now() here instead
         let timestamp = now();
         problems
             .iter()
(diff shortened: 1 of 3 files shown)


src/db/delivery_problems/mod.rs
```diff
@@ -64,6 +65,7 @@ where
     /// [limits]: ../settings/struct.BounceLimits.html
     pub fn check(&self, address: &EmailAddress) -> AppResult<()> {
         let problems = self.auth_db.get_bounces(address)?;
+        // TODO: When we start reading from the new datastore, use Utc::now() here instead
```
vladikoff: let's create a follow up issue if possible

src/db/delivery_problems/mod.rs
```diff
@@ -202,7 +207,10 @@ impl From<LegacyDeliveryProblem> for DeliveryProblem {
             address: source.address,
             problem_type: source.problem_type,
             problem_subtype: source.problem_subtype,
-            created_at: source.created_at,
+            created_at: Utc.timestamp(
```
vladikoff: there is a https://docs.rs/chrono/0.4.6/chrono/offset/trait.TimeZone.html#method.timestamp_opt as well, but hopefully we won't need it here for `source.created_at`

src/queues/mod.rs
```diff
@@ -177,8 +178,11 @@ impl Queues {
     fn record_complaint(&'static self, notification: &Notification) -> AppResult<()> {
         if let Some(ref complaint) = notification.complaint {
             for recipient in &complaint.complained_recipients {
-                self.delivery_problems
-                    .record_complaint(&recipient, complaint.complaint_feedback_type)?;
+                self.delivery_problems.record_complaint(
+                    &recipient,
+                    complaint.complaint_feedback_type,
+                    complaint.timestamp,
```
vladikoff: 👍

src/db/delivery_problems/mod.rs
philbooth: Oh right, yeah, didn't notice that this one panics. I'll switch.

philbooth:
Yeah, that dead code warning is stuff that was added in preparation for #185. I'm hoping to implement the rest of it tomorrow or Friday, then the warning will go away.

PR #236, opened 2018-11-08
refactor(api): reduce the number of different error kinds we return

### LINKED ISSUE #210: Tidy up the AppError
Having played with `failure` a bit over in syncstorage-rs, there's a few little changes I want to make to our `AppError` abstraction:

* Implement `Serialize` instead of exposing a `json` method, and update usage thereof.

* Don't violate the law of Demeter by returning the internal `AppErrorKind`, implement getters for access to specific properties instead.

* Remove some of the `AppErrorKind` variants that don't directly correlate with error states in our codebase (I mean generic ones like `BadRequest`, where we have specific validation errors to use instead, but not e.g. `InternalServerError` which still has uses).

* Have a cleaner hierarchy, with just a few kinds under the top-level `AppError` and then further sub-kinds under those where necessary (i.e. similar to what syncstorage-rs does in [src/error.rs](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/error.rs), [src/web/error.rs](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/web/error.rs) and [src/db/error.rs](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/db/error.rs)).

philbooth:
Fixes #210.

The error-handling in this repo grew pretty organically without us giving much thought to an over-arching strategy. The net result of that was a proliferation of different error types that weren't really adding much value.

The API here is simple and the 4xx responses are really limited to bad requests and bounce/complaint limit violations. Everything else is an internal server error. As such, the `error` module could be simplified by employing a blanket `Internal` error type to cover the multitude of niche errors that should never occur during normal conditions. The detail for those errors is still available in the `message` property and Sentry will still log the backtrace of course.

I also noticed that the Rocket catcher stuff wasn't really being used, so pulled that out too. We have a standard JSON error format like the rest of the FxA ecosystem, and we always want the response payload to be the serialisation of that structure.

One side-effect of all these changes is that the errno value has changed in most cases. However, I took care to preserve the errno for the bounce and complaint violations, because it's hard-coded in the auth server. It wouldn't have been that onerous to open a PR for commensurate changes over there, but the number of errors left here worked out in such a way that made sense not to bother.

It's a bit of a messy review, sorry. Lots of small, disparate changes with nothing meaty to sink your teeth into. The ultimate objective for it all was the slimmed-down `src/types/error/mod.rs`, which I'm happy with.

@mozilla/fxa-devs r?


src/api/send/test.rs
```diff
@@ -217,7 +208,8 @@ fn missing_to_field() {
     assert_eq!(response.status(), Status::BadRequest);
 
     let body = response.body().unwrap().into_string().unwrap();
-    let error: AppError = AppErrorKind::MissingEmailParams(String::from("")).into();
+    let error: AppError =
+        AppErrorKind::InvalidPayload(String::from("missing field `to` at line 7 column 5")).into();
```
philbooth: Oh, small bit of collateral improvement here: previously we didn't have rich validation detail in these 400 errors, now we do. Winning!

src/api/send/test.rs
```diff
@@ -231,7 +223,7 @@ fn missing_subject_field() {
         .header(ContentType::JSON)
         .body(
             r#"{
-      "to": [ "foo@example.com" ],
+      "to": "foo@example.com",
```
philbooth: And here, because we didn't have the error detail mentioned above, we hadn't actually noticed that a bunch of these tests were set-up wrong, and were failing for the wrong reason.

src/db/auth_db/mod.rs
```diff
@@ -124,15 +140,3 @@ impl DbUrls {
...
     }
 }
-
-impl From<UrlError> for AppError {
-    fn from(error: UrlError) -> AppError {
-        AppErrorKind::AuthDbError(format!("{}", error)).into()
-    }
-}
-
-impl From<RequestError> for AppError {
```
philbooth: I also discovered that a couple of these blanket conversions were causing the wrong error response. Here, any `reqwest::Error` would end up as an `AuthDbError`, then somewhere else there was a similar case of it (I forget where now).

src/db/delivery_problems/test.rs
```diff
@@ -113,15 +113,15 @@ fn check_soft_bounce() {
     assert_eq!(error.code(), 429);
     assert_eq!(error.errno().unwrap(), 107);
     assert_eq!(error.error(), "Too Many Requests");
-    assert_eq!(error.to_string(), "Email account soft bounced");
+    assert_eq!(error.to_string(), "Soft-bounce limit violated");
```
philbooth: I tweaked these error messages to make it clearer that the error is not saying an email necessarily bounced just now, but that the configured limits have been violated (so the bounce itself may have been some time ago).

src/types/error/mod.rs
```diff
@@ -139,199 +134,86 @@ impl<'r> Responder<'r> for AppError {
...
-            | AppErrorKind::InternalServerError => None,
+            AppErrorKind::Internal(_) => Some(100),
+            AppErrorKind::NotImplemented(_) => Some(101),
+            AppErrorKind::InvalidPayload(_) => Some(102),
+            AppErrorKind::InvalidNotification(_) => Some(103),
+            AppErrorKind::QueueError(_) => Some(104),
+            AppErrorKind::InvalidDuration(_) => Some(105),
+            AppErrorKind::Complaint { .. } => Some(106),
+            AppErrorKind::SoftBounce { .. } => Some(107),
+            AppErrorKind::HardBounce { .. } => Some(108),
```
philbooth: These last three are the ones that have been preserved to match their old values, in case you want to review that.

src/types/error/test.rs
```diff
@@ -2,58 +2,133 @@
...
 use super::*;
+use db::delivery_problems::{ProblemSubtype, ProblemType};
 
 #[test]
-fn bad_request() {
-    let error: AppError = AppErrorKind::BadRequest.into();
-    assert_eq!(error.code(), 400);
-    assert_eq!(error.error(), "Bad Request");
-    assert!(error.errno().is_none());
+fn internal() {
```
philbooth: Oh, one more small collateral improvement. Previously we didn't have test coverage in her for all our error types. Now we do.

src/types/error/mod.rs
```diff
@@ -344,66 +235,30 @@ impl AppErrorKind {
...
-
-#[catch(422)]
-pub fn unprocessable_entity() -> AppResult<()> {
-    Err(AppErrorKind::UnprocessableEntity)?
-}
-
-#[catch(429)]
-pub fn too_many_requests() -> AppResult<()> {
-    Err(AppErrorKind::TooManyRequests)?
+macro_rules! to_internal_error {
```
brizental: Nice!

PR #240, opened 2018-11-15
fix(ses): ensure html part gets charset=utf-8 content type

### LINKED ISSUE #239: Two differences in SMTP/MIME sent via email-service vs. auth-server, and encoding issue in Thunderbird
Not saying there's necessarily a bug here, but I inspected the email sent via email-service vs. one sent by auth-server and note these differences.

1. Via auth server has these two headers that email-service does not:
```
Content-Language: en
Sender: Firefox Accounts <accounts@firefox.com>
```

2. Via auth server, the HTML part is sent as:

```
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: quoted-printable
```

while with the email service it is:

```
Content-Transfer-Encoding: base64
Content-Type: text/html
```

philbooth:
Fixes #239.

The missing `charset` in the `Content-Type` header caused train 124 to fail. I could have patched it in 124 but it feels like we're close enough to train 125 to patch it there instead now (and I don't even know if there's ops bandwidth to test it out in 124 again anyway).

I've also added the `Sender` header, although I don't *think* its absence was causing problems. The spec seems pretty clear that it falls back to the value of `From` by default, but email servers are a capricious species so better to be explicit.

The `Content-Language` header can only be patched in the auth server (patch incoming). The `Content-Transfer-Encoding` header is by design.

@mozilla/fxa-devs r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: rust
 services:
   - redis-server
 cache: cargo
-branches:
-  only:
-    - master
 before_script:
   - rustup component add rustfmt-preview
 script:
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -6,7 +6,7 @@
 
 use std::{boxed::Box, collections::HashMap, convert::TryFrom};
 
-use emailmessage::{header::ContentType, Message, MessageBuilder, MultiPart, SinglePart};
+use emailmessage::{header::ContentType, Mailbox, Message, MessageBuilder, MultiPart, SinglePart};
 
 use self::{
     mock::MockProvider as Mock, sendgrid::SendgridProvider as Sendgrid, ses::SesProvider as Ses,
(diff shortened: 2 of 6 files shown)


philbooth:
> The `Content-Language` header can only be patched in the auth server (patch incoming).

Correction, it is this repo. I'll push another fix to this branch presently.

jrgm:
> I could have patched it in 124 but it feels like we're close enough to train 125 to patch it there instead now (and I don't even know if there's ops bandwidth to test it out in 124 again anyway).

I can put another train-124 stack train with this version of email-service tomorrow, but after I hear back  from SV about their plans for stage (since I'd have to temporarily move stage back to 124).

> but email servers are a capricious species so better to be explicit.

Indeed. Clients too.

philbooth:
> I can put another train-124 stack train with this version of email-service tomorrow, but after I hear back from SV about their plans for stage (since I'd have to temporarily move stage back to 124).

Hmm, meh. The main value to me for testing train 124 was to see if that panic was fixed before we got to train 125, which is the one that contains the important goodness for the OKR. Seems like we're fine on that front, so I'm cool with wrapping this in to train 125 (the test coverage is much better for these changes).

jrgm:
I wouldn't mind picking this change up tomorrow to move this forward and just get some live time in prod.

philbooth:
> I wouldn't mind picking this change up tomorrow to move this forward and just get some live time in prod.

Understood, I'll tag a 124 patch as well.

PR #246, opened 2018-11-20
chore(package): create an automated release script

### LINKED ISSUE #214: Automate the release process

philbooth:
Fixes #214.

Adds a shell script that automates most of the work involved in tagging releases. The only thing it doesn't do is `git push`, because a manual check that everything worked before publishing the tag is obviously a good thing.

The script has one slight advantage over what we're using elsewhere, in that the order of entries in the changelog will always be sane, even when we're tagging patch releases for older trains. That's because the step for updating the changelog inserts the summary immediately before the preceding tag, rather than always putting it at the top of the file regardless of where we are in the tree.

There are three other, small differences in the changelog:

* I removed the leading `v` from the version in the section headings (it's still present in the tag itself), to match the style in our other changelogs.

* The date will not be inserted for future releases, because I wasn't sure of a reliable, portable way to format the date across all Linux/Unix flavours. I figure it's not a big deal because the date and time show up in `git log` and on GitHub anyway.

* I removed the empty anchor tags because we don't need them/they don't even show up on GitHub (presumably they're there for the benefit of some other platform). I dislike them because they're kind of ugly for anyone viewing the file as text.

These are all of the steps that the script performs, in order:

1. Identify the last tag that precedes `HEAD`, assign it to `LAST_TAG`.

2. Get the list of intervening one-line commit messages from `git log $LAST_TAG..HEAD`.

3. Parse each commit message in turn, building up summary sections for the commit log. This includes an `Other changes` section for any commits that don't match one of the recrognised formats.

4. Split `$LAST_TAG` and bump the appropriate part, then assemble the bumped tag as `NEW_TAG`.

5. Update `Cargo.toml` and `Cargo.lock` with `$NEW_TAG`.

6. Insert the summary for `$NEW_TAG` into the changelog, immediately before the summary for `$OLD_TAG`.

7. Commit the changes.

8. Tag the release.

Lastly, I wonder if some of you might be unhappy that we've written this ourselves rather than using a 3rd-party tool. I get where you're coming from, but fwiw this script is almost a verbatim copy ([ref](https://gitlab.com/philbooth/please-release-me/blob/master/release.sh)) of something I've been using in all my side-projects for the last couple of years. So it's reasonably battle-hardened at this point, although I'm sure there are still lots of exciting bugs for us to discover when we run it on different machines to mine. :grin:

@mozilla/fxa-devs r?


RELEASE.md
```diff
@@ -1,28 +1,8 @@
 ## Creating Releases
 
-### Get dependencies if missing
+1. Run `scripts/release.sh` if you're doing a major train bump,
```
vladikoff: Thanks for updating this!

scripts/release.sh
```diff
@@ -0,0 +1,141 @@
...
+    ;;
+  *)
+    echo "Release aborted: Invalid argument \"$1\"."
+    exit 1
+    ;;
+esac
+
+while read -r COMMIT; do
+  HASH=`echo "$COMMIT" | cut -d ' ' -f 1`
+  MESSAGE=`echo "$COMMIT" | cut -d ':' -f 2- | awk '{$1=$1};1'`
```
vladikoff: does everyone have `awk`?

scripts/release.sh
```diff
@@ -0,0 +1,141 @@
...
+      REFACTOR_SUMMARY="$REFACTOR_SUMMARY\n* $AREA$MESSAGE ($HASH)"
+      ;;
+    "revert")
+      if [ "$REFACTOR_SUMMARY" = "" ]; then
+        REVERT_SUMMARY="### Reverted changes\n"
+      fi
+      REVERT_SUMMARY="$REVERT_SUMMARY\n* $AREA$MESSAGE ($HASH)"
+      ;;
+    *)
+      if [ "$OTHER_SUMMARY" = "" ]; then
```
vbudhram: We are probably inconsistent with it but other repos have a `chore` section. Not sure if worthwhile to have it in its own for put in other section.

philbooth: I figured `chore` makes sense to go under "Other changes", they'll mostly be `chore` items but occasionally other stuff might pop up in there too.

scripts/release.sh
rfk: I can confirm I have `awk` 😁

PR #256, opened 2018-11-29
refactor(settings): implement Default in enum_boilerplate!

### LINKED ISSUE #216: Convert string settings types to enums where possible

philbooth:
Fixes #216.

The last part of the enum refactoring, since all our enums have a default value we may as well implement `Default` in `enum_boilerplate!` too.

The other thing mentioned for #216 was to use `rusoto_core::Region` but actually it doesn't work so well for us. The serialised format is an array, `[ name, endpoint ]`, whereas we only want to specify the name in our settings. So I figure we can leave it like it is for now and decide whether we want to implement our own `AwsRegion` type later.

@mozilla/fxa-devs r?

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -7,7 +7,7 @@
 //! miscellaneous generally-used types.
 
 macro_rules! enum_boilerplate {
-    ($name:ident ($description:expr, $error:ident) {
+    ($name:ident ($description:expr, $default:ident, $error:ident) {
         $($variant:ident => $serialization:expr,)+
     }) => {
         #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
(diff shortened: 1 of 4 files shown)


src/types/mod.rs
```diff
@@ -25,6 +25,12 @@ macro_rules! enum_boilerplate {
             }
         }
 
+        impl std::default::Default for $name {
```
vladikoff: neat!

PR #258, opened 2018-12-05
wip: add an endpoint for writing configuration data

philbooth:
Related to #185.

Adds an endpoint that can write (or merge) config data to Redis. There's quite a lot here for a single reviewer, hence opening as `WIP` and I'll pull out some smaller chunks that can be reviewed independently over the next couple of days.


src/api/config/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use db::config::{Config as ConfigDb, Data as ConfigData};
+use logging::MozlogLogger;
+use types::error::{AppError, AppErrorKind, AppResult};
+
+/// Payload for `POST /config`.
+#[derive(Debug, Deserialize)]
+struct Payload {
+    /// Flag indicating whether this payload clobbers existing config or should be merged with it.
+    clobber: Option<bool>,
```
vbudhram: It might be a little clearer if it was called `replace` and noted that default is to merge configs.

src/api/config/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+            })
+            .map(|json| json.into_inner())
+    }
+}
+
+#[post("/config", format = "application/json", data = "<payload>")]
+fn handler(
+    payload: AppResult<Payload>,
+    config_db: State<ConfigDb>,
+    // TODO: do we need the logger?
```
vbudhram: If it isn't too hard, I think it might be useful.

src/api/config/test.rs
```diff
@@ -0,0 +1,114 @@
...
+#[test]
+fn request_with_all_the_things() {
+    let client = setup();
+
+    let mut response = client
+        .post("/config")
+        .header(ContentType::JSON)
+        .body(
+            r#"{
+              "clobber": true,
```
vbudhram: Are these values documented somewhere?

src/api/config/test.rs
```diff
@@ -0,0 +1,114 @@
...
+                    { "period": "2 years", "limit": 10 }
+                  ],
+                  "soft": [
+                    { "period": "3 minutes", "limit": 1 }
+                  ]
+                },
+                "queue": "https://sqs.us-east-1.amazonaws.com/1234567890/blee",
+                "rules": [
+                  { "percentage": 50, "precedence": -127, "provider": "sendgrid" },
+                  { "percentage": 100, "precedence": 0, "provider": "socketlabs", "regex": "^socketlabs@mozilla\\.com$" },
```
vbudhram: Should the highest precedence be 0? It does seem strange that the highest precedence would be negative in this case.

src/api/config/test.rs
```diff
@@ -0,0 +1,114 @@
...
+        .dispatch();
+
+    assert_eq!(response.status(), Status::Ok);
+
+    let body = response.body().unwrap().into_string().unwrap();
+    assert_eq!(body, json!({}).to_string());
+}
+
+#[test]
+fn request_without_optional_fields() {
```
vbudhram: Can add more tests for clobber.

